PR #10, opened 2016-11-24
Added some models and services

adrianhopebailie:
Moved to lots of classes to org.interledger.core.ilp.ledger package


adrianhopebailie:
Thanks @sappenin - missed the review somewhere!

PR #12, opened 2017-01-23
Add FulfillTransfer and refactoring for SPSP and ILQP

adrianhopebailie:


src/main/java/org/interledger/setup/model/Receiver.java
```diff
@@ -0,0 +1,18 @@
...
+
+import org.interledger.ilp.InterledgerAddress;
+
+public interface Receiver {
+  
+  InterledgerAddress getAccount();
+  
+  CurrencyUnit getCurrencyUnit();
+
+  int getPrecision();
```
andrew-g-za: If we are strongly typing into CurrencyUnit, is the precision and scale still meaningful? For precision at least, isnt CurrencyUnit.getDefaultFractionDigits equivalent?

src/main/java/org/interledger/setup/model/ReceiverQuery.java
```diff
@@ -0,0 +1,5 @@
+package org.interledger.setup.model;
+
+public interface ReceiverQuery {
+
```
andrew-g-za: shouldnt we define the receiver query to contain the URI of the receiver endpoint, or do you forsee SPSP supporting alternative receiver endpoints?

src/main/java/org/interledger/setup/spsp/model/Invoice.java
```diff
@@ -0,0 +1,15 @@
+package org.interledger.setup.spsp.model;
```
andrew-g-za: shouldnt this just be org.interledger.setup.model? Are we expecting other setup models apart from SPSP?
for brevity, same comment applies to the other classes in this package (InvoiceStatus, Payee etc).

src/main/java/org/interledger/setup/spsp/model/Payee.java
```diff
@@ -0,0 +1,13 @@
+package org.interledger.setup.spsp.model;
+
+import java.net.URI;
+
+public interface Payee extends SpspReceiver {
+  
+  public String getName();
+
+  //TODO Should we return a java.awt.Image?
```
andrew-g-za: no, i would definitely leave it as a url - there's no point in fetching that data unless someone explicitly asks for it.

src/main/java/org/interledger/ilp/exceptions/InterledgerQuotingProtocolException.java
```diff
@@ -1,4 +1,4 @@
-package org.interledger.ilp.core.exceptions;
+package org.interledger.ilp.exceptions;
 
 public class InterledgerQuotingProtocolException extends Exception {
```
adrianhopebailie: Could do. The whole architecture of the Exceptions needs some review now that the overall architecture is stabilizing (kind of)

src/main/java/org/interledger/setup/SetupService.java
```diff
@@ -0,0 +1,40 @@
...
+import org.interledger.setup.model.ReceiverQuery;
+
+
+public interface SetupService {
+
+  /**
+   * The sender (this service) queries the receiver endpoint to get information about the type of 
+   * payment that can be made to this receiver:
+   *
+   * @param receiverEndpoint
```
adrianhopebailie: Will do. As with the Exceptions. JavaDoc needs a complete review. Will put that on my TODO list.

src/main/java/org/interledger/setup/model/Receiver.java
adrianhopebailie: Hmmm, good point. Need to think about this carefully because it could be that we need a custom factory for CurrencyUnit.

It currently uses the default and fetches the CurrencyUnit based on currency code. This means it's possible that the system values for precision and scale are different to those supported by the ledger.

I suggest we leave that in for now and I'll work on a complete solution, possibly involving a custom CurrencyUnit factory.

src/main/java/org/interledger/setup/model/ReceiverQuery.java
adrianhopebailie: The idea here is to define SPSP as one of potentially many setup protocols. So a receiver does not necessarily have a URI (this is defined in SpspReceiver).

Different setup protocols will have different ways to return a Receiver so another SetupService might actually just build it locally.

src/main/java/org/interledger/setup/spsp/model/Invoice.java
adrianhopebailie: As above, yes we are. See: https://github.com/interledger/rfcs/issues/43

src/main/java/org/interledger/setup/spsp/model/Payee.java
adrianhopebailie: Agree

PR #20, opened 2017-03-29
Added OER Support, restored from ancient java-crypto-condition version

earizon:
Restored (org.interledger.encoding.)OerInputStream and OerOutputStream

### DIFF
--- a/src/main/java/org/interledger/encoding/OerInputStream.java
+++ b/src/main/java/org/interledger/encoding/OerInputStream.java
@@ -1,7 +1,6 @@
 package org.interledger.encoding;
 
 import java.io.EOFException;
-
 import java.io.IOException;
 import java.io.InputStream;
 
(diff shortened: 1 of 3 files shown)


earizon:
@sappenin 
> Any chance we could get unit tests for these? There is a new ILP binary packet being defined that might be a good candidate, but perhaps a more general set of tests is in order?

I just added a "draft" unit-test that writes OER to memory, then reads it backs. Far away from perfect, but still useful as documentation. Definetly testing with real ILP packet data is a "must".

https://github.com/interledger/java-ilp-core/pull/20/files#diff-287749664ce9c22151c7148d5b068c79

sappenin:
LGTM!  Can you merge this, or do I need to?

adrianhopebailie:
@earizon, you should be able to merge. Let me know if not.

adrianhopebailie:
@earizon the flow we are adopting is for committers to merge their own PRs once they have been reviewed and approved.

I've merged this one so it doesn't go stale

earizon:
@adrianhopebailie  (overflowed by a jungle of emails). Got it for next time!

PR #21, opened 2017-04-04
Earizon refactoring exceptions

earizon:
Refactor ILP Exceptions according to the latest RFC:
 https://github.com/interledger/rfcs/blob/master/0003-interledger-protocol/0003-interledger-protocol.md

I noticed the existing Exception/s use a hierarchy InterledgerException <-> child classes, but this could be overkilling now since there can be (pottencially) tens or hundreds of differnt errors and the public Enum InterledgerException.ErrorCode serves to have a compliant implementation with the RFCs. Also, it's not expected for code to extend the behaviour and/or data of the base exception (most probably the code will limit to serialize and send back the wire to other connectors/ledger/clients)

I leave as TODO comments some doubts I still have:
-  Must ILQP Exception be treated as an "embeded" exception inside the InterledgerException.data ?
- how the exception be serialized?


src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
```diff
@@ -1,36 +1,137 @@
...
+import java.util.ArrayList;
+import java.time.ZonedDateTime;
+
+import org.interledger.ilp.InterledgerAddress;
+
 /**
  * Base ILP exception.
  *
- * @author Manuel Polo [mistermx@gmail.com]
+ * RFC REF: https://github.com/interledger/rfcs/blob/master/0003-interledger-protocol/0003-interledger-protocol.md
```
adrianhopebailie: We should point to the public URL: https://interledger.org/rfcs/0003-interledger-protocol/#errors

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
```diff
@@ -1,36 +1,137 @@
...
+     *    be relatively big (data is up to 8192 bytes) and cloning the full exception
+     *    to add a single a member to the list could cause performance/memory trouble in
+     *    some scenarios.
+     */
+    private final ArrayList<InterledgerAddress> forwardedBy;
+    private final ZonedDateTime triggeredAt;
+    private final String data;
+    /**
+     * Constructs an instance of <code>InterledgerException</code> 
+     * Check the RFC https://github.com/interledger/rfcs/blob/master/0003-interledger-protocol/0003-interledger-protocol.md
```
adrianhopebailie: We should point to the public URL: https://interledger.org/rfcs/0003-interledger-protocol/#errors

src/main/java/org/interledger/ilp/ledger/LedgerAdaptor.java
```diff
@@ -66,7 +66,7 @@
    * @param transfer The transfer being proposed.
    * @param reason The reason why the transfer should be rejected.
    */
-  void rejectTransfer(LedgerTransfer transfer, TransferRejectedReason reason);
+  void rejectTransfer(LedgerTransfer transfer, InterledgerException.ErrorCode reason);
```
adrianhopebailie: Why not pass in the whole exception?

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
```diff
@@ -1,36 +1,137 @@
...
+            }
+            code = code.trim();
+            name = name.trim();
+            if (code.length()!=3) {
+                throw new RuntimeException("error code length must be equal to 3");
+            }
+            char type = code.charAt(0);
+            if (type != 'F' && type !='T' && type !='R'){
+                throw new RuntimeException("error code must be := 'F' | 'T' | 'R'");
+            }
```
adrianhopebailie: We should expose ErrorType as an enum too (and as a property of the error)

src/main/java/org/interledger/ilp/ledger/LedgerAdaptor.java
earizon: @adrianhopebailie Thanks for the review. I just updated the code. 
- Regarding the ErrorType it's exposed now as a (enum) property of the error throught the getErrorType, but at Exception-creation-time it's the first character of errorCode that will used/parsed to set the type since developers will have to indicate the errorCode and forcing them to also indicate the errorType will just cause them to set duplicated information (with marginal performance benefits) and the possibility of coding errors.

Let me know if you have a different point of view.

- Note the TODO:(RFC) indicates some point that could be better to fix (or better explain) on the RFC.

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
```diff
@@ -1,36 +1,207 @@
...
+        // TODO:(RFC) ADD FORBIDEN ERROR IF NO CREDENTIALS ARE PROVIDED OR LAUNCH "T01 LEDGER-UNREACHABLE" ???
+        // IMAGINE node1 -> conector1 -> conector2 -> (ERR. FORBIDEN) contector3
+        //  - How must routing be updated in  conector1/2/3 ?
+
+        public  final ErrorType errorType;
+        private final String code;
+        private final String name;
+
+        private ErrorCode(String code, String name) {
+            if (code==null || name==null){
```
sappenin: Not a big enough deal to change, but we should prefer `Objects.requireNonNull` when checking a precondition.  It's easier to read, and will allow you to throw a discrete RuntimeException so the caller knows which value was null in the error.

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
```diff
@@ -1,36 +1,207 @@
...
+                throw new RuntimeException("code and/or name can not be null");
+            }
+            code = code.trim();
+            name = name.trim();
+            if (code.length()!=3) {
```
sappenin: Will this always be the case?  Will there perhaps be custom error codes, which might have length > 3?

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
```diff
@@ -1,36 +1,207 @@
...
+
+    public String getData() {
+        return data;
+    }
+
+    /**
+     * Used by connectors to add themself to the forwaredBy list
+     * @param selfAdd connector ILP address.
+     */
+    public void addSelfToForwardedByList(InterledgerAddress selfAdd) {
```
sappenin: Perhaps add a note here to indicate that this is _not_ threadsafe?

From the Javadoc of ArrayList:

> <p><strong>Note that this implementation is not synchronized.</strong>
> If multiple threads access an <tt>ArrayList</tt> instance concurrently,
> and at least one of the threads modifies the list structurally, it
> <i>must</i> be synchronized externally.  (A structural modification is
> any operation that adds or deletes one or more elements, or explicitly
> resizes the backing array; merely setting the value of an element is not
> a structural modification.)  This is typically accomplished by
> synchronizing on some object that naturally encapsulates the list.
>
> If no such object exists, the list should be "wrapped" using the
> {@link Collections#synchronizedList Collections.synchronizedList}
> method.  This is best done at creation time, to prevent accidental
> unsynchronized access to the list:
> <pre>List list = Collections.synchronizedList(new ArrayList(...));</pre>

...Or we could wrap the EMPTY_ForwardedBy_List at constructions time?

src/main/java/org/interledger/ilp/exceptions/InterledgerQuotingProtocolException.java
```diff
@@ -2,6 +2,19 @@
 
 public class InterledgerQuotingProtocolException extends Exception {
```
sappenin: Should this extend InterledgerException?  Or is this a distinct thing...?

src/main/java/org/interledger/ilp/ledger/model/TransferStatus.java
```diff
@@ -28,5 +28,15 @@ public static TransferStatus valueOf(int status) {
         throw new IllegalArgumentException("Invalid TransferStatus status " + status);
     }
   }
+  
+  public static TransferStatus parse(String status) {
+      status = status.toLowerCase();
+      if ("proposed".equals(status) || "0".equals(status)) { return PROPOSED; }
```
sappenin: Should we be looking for the ordinal value here?  Seems like that should be invallid.

src/main/java/org/interledger/ilp/ledger/model/TransferStatus.java
```diff
@@ -28,5 +28,15 @@ public static TransferStatus valueOf(int status) {
...
+      if ("prepared".equals(status) || "1".equals(status)) { return PREPARED; }
+      if ("executed".equals(status) || "2".equals(status)) { return EXECUTED; }
```
sappenin: `String#equalsIgnoreCase`?

sappenin:
@earizon, you asked:

> how the exception be serialized?

Do you mean Java serialization (i.e., java.io.Serializable) or do you mean via some encoding like JSON, Protobuf, etc?

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
adrianhopebailie: I think this should only deal with Interledger Error codes which will always be 3 chars. Custom/application errors will need to extend this or have this as a `cause`.

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
adrianhopebailie: Why not just use an array? Seems like a lot of unnecessary overhead to use a List

src/main/java/org/interledger/ilp/exceptions/InterledgerQuotingProtocolException.java
adrianhopebailie: I think this is distinct

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
earizon: fixed in last commit

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
earizon: The RFC (https://interledger.org/rfcs/0003-interledger-protocol/#errors) is clear about it:
code IA5String (SIZE(3))

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
earizon: Argg Threads!!! I always forget about threads! (I miss single-threaded scripting code now)
I just opted for the simple secure probe solution. Use an inmutable InterledgerAddress[] inmutable object.

I notice (correct me if I'm wrong) that there will be two scenarios:
- A node (ledger/connector/...) raise a new exception. In that case forwaredBy will be an empty array.
- A connector receives the exception, add itself  to the forwardedBy list and send back to other ILP nodes.  In this case I added an static utility method addSelfToForwardedBy to "easify" the use of the constructor.
 As a bonus this helper/utility static method check for "loops" (the connector find himself already added to the forwaredBy list) and throws a RuntimeException.

Not 100% sure what to do when detecting a loop but launching an exception is safer than ignoring it. This probably is a sympton of something really wrong in the routing setup.

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
```diff
@@ -1,36 +1,196 @@
...
+        @Override
+        public String toString(){
+            return this.code + "-" +this.name;
+        }
+    }
+
+    private final ErrorCode errCode;
+    private final InterledgerAddress triggeredBy;
+
+    private final InterledgerAddress[] forwardedBy;
```
sappenin: I prefer not using Arrays for a variety of reasons.  See [here](http://stackoverflow.com/questions/6100148/collection-interface-vs-arrays) or [here](http://eclipsesource.com/blogs/2014/04/11/3-good-reasons-to-avoid-arrays-in-java-interfaces/) for some of them.  If you want to use an Array internally, I would be ok with that, but our interface (i.e., the constructor, etc) should use a Collection.

Thoughts?

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
```diff
@@ -1,36 +1,196 @@
...
+        return triggeredAt;
+    }
+
+    public String getData() {
+        return data;
+    }
+
+    // Utility static methods 
+
+    public static InterledgerAddress[] addSelfToForwardedBy(InterledgerAddress[] forwardedBy, InterledgerAddress selfAddress) {
```
sappenin: In general I don't like static methods like this.  It leaks an implementation detail into the formal interface of the Exception class.  Just because this class requires an Array of addresses (or, hopefully, a Collection -- see comment above) doesn't mean it needs to have any business helping callers create an Array.  

Instead, your previous implementation was better -- you take an InterledgerException, and you add a new address to it (use Collections to synchronize if desired, or return a new Immutable Exception that contains an immutable collection of addresses).

src/main/java/org/interledger/ilp/exceptions/InterledgerQuotingProtocolException.java
earizon: From the RFC: "Protocols built on top of ILP that define behavior for certain errors SHOULD specify the encoding format of error data."
I understand that an InterledgerException must be thrown and the data field use to add the ILQP specific data. I'm not sure at all. Maybe the JS code is a good reference point?

earizon:
@sappenin 
>Do you mean Java serialization (i.e., java.io.Serializable) or do you mean via some encoding like JSON, Protobuf, etc?
I mean the final encoding. I think is well defined now on the RFC, but I had no time to check it properly. 
I think it must be "symetrical" to the ILP packet and must be OER encoded before sending it back in any transport (HTTP, WebSocket, ...). Again not 100% sure.

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
sappenin: I would advocate that in the connector case, the constructor should allow the Connector to take an existing Exception, plus a new address, and construct an entirely new Exception from there.  That way, everything inside of the Exception can be final/immutable.

So, maybe something like this:

```
public InterledgerException(final InterledgerException interledgerException, final   InterledgerAddress triggeredBy) {
    // ... copy everything from interledgerException, and internally add triggeredBy to the list of addresses.
}
```

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
earizon: @sappenin This is a old debate. 
Must we use interfaces for *everything* and then implement them in concrete clases?

My point of view is that interfaces must be used to describe behaviour whose implementation can vary for different reasons (security, performance, memory, ...) while concrete classes must be used to describe data-centric (DTO) objects. Class instances could be interpreted as "rows" in a database table. (In the case of exception to lines of logs in syslog maybe). If the RFC mandates an object that must have some well defined fields (errCode that is 3 chars long followed by an InterledgerAddress ...) then defining an interface will not help at all. It will promote "random" implementations of a concrete schema that was well defined from the scratch.  Different implementation are not expected to improve the code. A final developer will expect to have a "box" (class) where he can place the data. This is similar to the java.security.spec classes. Is not expected for Java Security implementators to reimplement those "data objects". (but it expected that many different ways to fill them exists).

sappenin: @earizon Sorry for the confusion --  when I wrote "interface" I meant it in a general sense, not in the formal "Java Interface" sense.  Perhaps I should use the word "contract" instead.  We're advertising a contract when we write this Exception class, and we need to be careful about what that contract is.

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
earizon: @sappenin Done! Yes this is a trouble-free solution. Let me now what you think of the last commit!

I noted while coding that, since InterledgerAddress is defined as an interface there is a risk now of creating code like:
   address1.equals(address2); that can fail since "equals" can not be defined for concrete instances and default to the reference comparation (both objects points to the same memory address)
  I'm forced to do:
  address1.getValue().equals(address2.getValue())
This is an example of why interfaces can be problematic for "data" objects.

src/main/java/org/interledger/ilp/exceptions/InterledgerException.java
adrianhopebailie: Suggest using `List<InterledgerAddress>` then?

sappenin: If ordered, yes.  If not, then `Collection`.  Can there be dups?  If not, then maybe a `Set`?

adrianhopebailie: My assumption was this is ordered. There could be dups so I don't think `Set`

sappenin:
I spent some more time thinking about this PR last night, working through some of the implementation's constructor cases for when a developer might use this class.

It occurred to me that we should consider splitting this class into two classes: One, an **Interface** that defines the contract for what's defined in RFC-3 (I'll call that `InterledgerError.java` for now); and Two, a **class** that actually extends RuntimeException (I'll call that `InterledgerException.java`).

In this potential new structure, `InterledgerException.java` could provide constructors that mirror those in `RuntimeException` and/or `Throwable`, in addition to taking a single extra property of type `InterledgerError`.  

My rationale for this change is that these two classes will be used in very different contexts.

* `InterledgerError.java` is a Java representation of the error object being passed around on the wire in some format -- likely JSON or Protobuf, or some standardized ASN.1 OER encoding.  This object (an Interface so developers can make their own implementations) should have no knowledger of things like a `Throwable` or a `message` or anything else in the Java exception heirarchy -- none of these details should leak into serialization formats (like JSON, OER, etc).  Think of this as an object that models ILP base objects in the same way InterledgerAddress.java does.

* `InterledgerException.java`, on the other hand, is a formal Java exception (extending `RuntimeException`) that developers can "throw" and "catch" in Java code and has a  "has-a" relationship with `InterledgerError` (meaning, an `InterledgerException` has an optionally-present `InterledgerError`)

In this way, Java programs can throw InterledgerExceptions all day long, handle them per-usual, but when it comes time to actually send something "on the ILP wire", like to a Connector, then the `InterledgerError` interface should be used.

Thoughts?

earizon:
@sappenin From the RFCs looks to me that errors are always considered exceptions, since they are supposed to stop normal flow, trigger a different message response and message format to clients, and they are añsp supposed to be treated in a completly different way to standard non-exception flow controls. I think the best (or the most conservative) approach would/will be to do what the JS reference implementation does. I couldn't find any related code in the interledgerjs repos yet.

sappenin:
> From @earizon: From the RFCs looks to me that errors are always considered exceptions, since they are supposed to stop normal flow, trigger a different message response and message format to clients, and they are supposed to be treated in a completely different way to standard non-exception flow controls.

@earizon No disagreements from me about InterledgerErrors being exception-cases.  Instead, I was referring to a difference between modeling the ILP-layer and the Java exception layer, and that we probably want those two layers to be different classes.  

For example, consider this typical API flow:

**Hypotehtical HTTP API**
1. Internal code throws a Java `Exception`.
1. Internal code catches the exception, looks at it's type or message, or whatever, and decides to map that to an HTTP status code and JSON payload.
1. API emits JSON that looks like this:
```
{
  "code": 400,
  "message": "You can't do that because we don't allow it"
}
```

Notice that the payload emitted by the API in this "exception case" doesn't include anything from the `java.lang.RuntimeException` class heirarchy.  There's no `message` or `stackTrace` or `cause` or `suppressedExceptions` or anything like that.  So, in my mind, we should model both things - 1.) the payload that will actually be returned to an ILP client;  and 2.) An exception class that can be used in Java code to try/catch.

To illustrate that using an Interledger example, consider a flow like this (again, this is just an example to illustrate a point):

**Hypothetical ILP Ledger API**
1. Internal code tries to execute a transfer, but the amount is negative.
1. Internal API throws a `MyCoolInvalidAmountException`.  This will have an internal `stackTrace`, potentially a `message`, and let's say some internal attributes that should not be exposed to the outside world.
1. An exception-mapper catches the `MyCoolInvalidAmountException`, and then needs to marshal it to a payload that looks like what's defined in RFC-3.  It transforms `MyCoolInvalidAmountException` into an instance of `InterledgerError` (see my previous comment), and returns this to the API layer.
1. API emits JSON that looks like this:
```
{
  "code":"F03",
  "name":"Invalid Amount",
  "triggeredBy":"g.foo.bar",
  ... // other RFC-3 fields.
}
```

Key point to notice is that the final payload has no mention of anything Java-Exception-related, because it's not a Java Exception.  It's something else.  And because it's something else, it should not extend from anything related to `java.lang.Throwable`.  

This will allow any implementor to use the Java-Exception we provide (`InterledgerException.java`) or create their own exceptions -- but ultimately, return a standard ILP payload object (`InterledgerError.java`).

Thoughts?

sappenin:
All, to illustrate what I was talking about in my previous comments, I submitted an alternate PR [here](https://github.com/interledger/java-ilp-core/pull/23).  

While I do prefer my split-design, I am open if the group feels otherwise, so we can certainly close my PR in favor of this design.  

Sometimes illustrating with code is easier to explain than written words trying to explain code.

src/main/java/org/interledger/ilp/exceptions/InterledgerQuotingProtocolException.java
sappenin: @earizon & @adrianhopebailie: I submitted [IL-RFC-#168](https://github.com/interledger/rfcs/issues/188) to remove the `data` field in the `InterledgerError`, or else clarify its usage.

IMHO, it doesn't make sense to include a field that _could_ mean anything (ILQP info, or just debug info, or nothing).  Instead, if sub-protocols are going to specify data that is added into an `InterledgerError`, then they should be specifying it.

At any rate, would appreciate any feedback, here or in that ticket.

earizon:
@sappenin I noticed you are thinking about a different way of handling exception to the one I had in mind:
> Internal code tries to execute a transfer, but the amount is negative.
> "An exception-mapper catches the MyCoolInvalidAmountException...  It transforms MyCoolInvalidAmountException into an instance of InterledgerError "

I see we diverge in the way exceptions must be handled. 

The way I see it, the code must not create a MyCoolInvalidAmountException, then later on map it to an InterledgerError. 
I think It's much safer (and IMHO, simpler) that the code detecting the error (amount is negative) will be forced to use a well defined InterledgerError at throw time. Allowing developers to be "lazy" when launching exceptions can produce buggy code. As Adrian once said "protect developers from themselves". Don't let them do something as 
"launch it know, think later how to map it to an ILP error" 
or 
"something/someone will (magically) know how to handle this exception and what it must do with it".

 If the API just let developers report exceptions in an unique way (throw an InterledgerException with well defined InterlederErrors provided by the RFCs or ask someone else with more experience what to do) there will not be missunderstanding about how to use it.

The main controller listening for incomming requests must be limited to serialize exceptions "comming back" in the catch(InterledgerException e) block and send them back to previous nodes. It must have NO "mapping" logic or allow different implementations to "interpret" the flow of ILP errors.
 Different implementations can choose where is the best place to place the "catch(InterledgerException e)" (a custom dispatcher, a common controller, a filter,...), the details inside the catch block (logging, ...) before serializing and sending it back to clients.

Other non-ILP related exceptions must be procesed in any proper way by the implementation and finally re-launch a well-defined ILP Error "T00 Internal Error".

adrianhopebailie:
@sappenin, @earizon sorry I got a little lost in PRs and email threads on this and didn't have a chance to dig into the details. 

I also hadn't realized @sappenin had proposed another PR and logged a number of related issues too.

@earizon  Was the consensus that this was the best solution or did you modify it "against your will" 😄 

My thinking is; let's use this as is for now and start building some implementations. We'll quickly discover if the design we have settled on is good or needs to change.

earizon:
@adrianhopebailie I have not a hard opinion. Spliting in Error and Exception makes the code slightly more readable and clean anyway.

PR #24, opened 2017-04-11
First attempt at InterledgerPacket

sappenin:
This PR is a first-attempt at the InterledgerPacket and ILQP Packet interfaces.  The ILQP interfaces are not complete, but are instead stubs to help us determine the best design for these classes.

**Design Questions**
- There are similarities between the overall InterlederPacket, and its children (InterledgerPaymentPacket.java and InterledgerQuotingProtocolPacket.java).  This PR unifies them, but perhaps there's an argument in favor of making them completely separate classes in the Java heirarchy.
 - We should create some concrete implementations of the interfaces in an attempt to see how accurate/desirable the interface design is.  I plan to do that later this week, but I encourage others to do the same.

**Design Note**
This PR chooses to model packets as interfaces instead of concrete classes.  This is to allow developers to choose implementations of these packets using different styles and/or technologies.  For example, @sappenin might create implementations based upon the [Immutables Project](https://immutables.github.io), while others might create implementations using other libraries like [Lombok](https://projectlombok.org/) or even Java POJOs.  

To this end, [java-ilp-core](https://github.com/interledger/java-ilp-core) should be flexible enough to accommodate various technologies chosen by implementors, while at the same time providing default implementations to aid developers who may simply want something "out of the box" -- see [InterledgerAddressBuilder.java](https://github.com/interledger/java-ilp-core/blob/feature/updated-ilp-packet/src/main/java/org/interledger/ilp/InterledgerAddressBuilder.java) for one example of how this project might accomplish that.


earizon:
I think it doesn't match the specs at https://github.com/interledger/rfcs/tree/master/asn1

There is an InterledgerPacket containing the type and the specific packet for each protocol
(Payment, PaymentRequest, Quote*Resquest, Quote*Response).

I think using stream readers/writers (extending the OER reader/writer) will simplificate the code.
As I see it protocol paquets must not be aware of the InterledgerPacket (must not extends it).

I send a summary of the specs at https://github.com/interledger/rfcs/tree/master/asn1
for clarity:

```
PacketSet PACKET ::= { {1 InterledgerProtocolPaymentMessage} | {2 QuoteLiquidityRequest} |  ... }
-------------------------------------------+----------------------------------------
InterledgerPacket ::= SEQUENCE {           | InterledgerProtocolError ::= SEQUENCE {
    -- One byte type ID                    |     code IA5String (SIZE (3)),
    type PACKET.&typeId ({PacketSet}),     |     name IA5String,
    -- Length-prefixed header              |     triggeredBy Address,
    data PACKET.&Type ({PacketSet}{@type}) |     forwardedBy SEQUENCE OF Address,
}                                          |     triggeredAt Timestamp,
                                           |     data OCTET STRING (SIZE (0..8192))
                                           | }
-------------------------------------------+----------------------------------------


-------------------------------------------+----------------------------------------
InterledgerProtocolPayment ::= SEQUENCE {  | InterledgerPaymentRequest ::= SEQUENCE {
    amount UInt64,                         |     version UInt8, -- IPR Version
    account Address,                       |     condition UInt256, -- Cryptographic condition
    data OCTET STRING (SIZE (0..32767)),   |     packet OCTET STRING -- ILP Payment Packet
    extensions SEQUENCE {  ... }           | }
}                                          |                                         
-------------------------------------------+----------------------------------------

QuoteLiquidityRequest ::= SEQUENCE {           QuoteLiquidityResponse ::= SEQUENCE {
    requestId UInt128,                             requestId UInt128,
    destinationAccount Address,                    liquidity LiquidityCurve,
    destinationHoldDuration UInt32,                appliesToPrefix Address,
    extensions SEQUENCE { ...  }                   sourceHoldDuration UInt32,
}                                                  expiresAt Timestamp,
                                                   extensions SEQUENCE { ...  }
                                               }

QuoteBySourceAmountRequest ::= SEQUENCE {      QuoteBySourceAmountResponse ::= SEQUENCE {
    requestId UInt128,                             requestId UInt128,
    destinationAccount Address,                    destinationAmount UInt64,
    sourceAmount UInt64,                           sourceHoldDuration UInt32,
    destinationHoldDuration UInt32,                extensions SEQUENCE { ...  }
    extensions SEQUENCE { ...  }               }
}

QuoteByDestinationAmountRequest ::= SEQUENCE {  QuoteByDestinationAmountResponse ::= SEQUENCE {
    requestId UInt128,                              requestId UInt128,
    destinationAccount Address,                     sourceAmount UInt64,
    destinationAmount UInt64,                       sourceHoldDuration UInt32,
    destinationHoldDuration UInt32,                 extensions SEQUENCE { ...  }
    extensions SEQUENCE { ...  }                }
}
```

src/main/java/org/interledger/ilp/packets/InterledgerPaymentPacket.java
```diff
@@ -0,0 +1,56 @@
...
+ * Payment Packet matches the amount actually delivered by the transfer. The receiver decodes the
+ * data of the ILP Payment Packet and matches the condition to the packet. The receiver MUST
+ * confirm the integrity of the ILP Payment Packet, for example with a hash-based message
+ * authentication code (HMAC). If the receiver finds the transfer acceptable, the receiver releases
+ * the fulfillment for the transfer, which can be used to execute all prepared transfers that were
+ * establlished prior to the receiver accepting the payment.</p>
+ */
+public interface InterledgerPaymentPacket<D> extends InterledgerPacket<IlpPaymentPacketType, D> {
+
+  /**
```
sappenin: This should include the `amount` and `account` accessors...

src/main/java/org/interledger/InterledgerPacket.java
```diff
@@ -0,0 +1,41 @@
...
+import org.interledger.ilp.InterledgerAddress;
+
+/**
+ * <p>Interledger Packets are attached to Interledger transfers to facilitate Interledger
+ * payments.</p>
+ *
+ * @param <T> The generic type of this packet, as defined by implementations.
+ * @param <D> The generic type of the data payload, as defined by implementations.
+ */
+public interface InterledgerPacket<T extends InterledgerPacketType, D> {
```
sappenin: Per @earizon's comment in the main "discussion" area, this should probably be as follows:

```
public interface InterledgerPacket<T extends InterledgerPacketType, D> {
 T getType();
 D getEncodedData();
}
```

The methods `getDestinationAmount` and `getDestinationAccount` should be moved down into `InterledgerPaymentPacket.java`.

src/main/java/org/interledger/InterledgerPacketType.java
```diff
@@ -0,0 +1,42 @@
...
+  Integer ILQP_QUOTE_BY_SOURCE_AMOUNT_RESPONSE_TYPE = 5;
+  Integer ILQP_QUOTE_BY_DESTINATION_AMOUNT_REQUEST_TYPE = 6;
+  Integer ILQP_QUOTE_BY_DESTINATION_AMOUNT_RESPONSE_TYPE = 7;
+
+  /**
+   * The packet's type identifier, as specified by IL-RFC-3.
+   *
+   * @return An {@link Integer} representing the type of this packet.
+   */
+  Integer getTypeIdentifier();
```
adrianhopebailie: Why `Integer` as opposed to `int`?

src/main/java/org/interledger/InterledgerPacketType.java
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * A URI representing the formal type of this packet per the Interledger Header Type registry
+   * maintained at IANA.
+   *
+   * @return An instance of {@link String}.
+   * @see "http://www.iana.org/assignments/interledger-header-types"
+   */
+  String getTypeUri();
```
adrianhopebailie: Return a URI?

adrianhopebailie:
Re-reading @earizon 's comments I think I agree with him. We should maybe have some simple business objects that represent the packets and leave complexity like mapping between type ids etc to readers/writers.

I'd go so far as to define the `data` member of an `InterledgerProtocolPayment` class as simply a `byte[]`.

src/main/java/org/interledger/InterledgerPacketType.java
sappenin: No major reason except using `Integer` allows the value to be `null` in Java, which for JSON deserialization purposes could be useful, whereas using an 'int' would default to 0 in Java, which would not be accurate in JSON.

sappenin:
I'm going to close this PR - the feedback in here was great, so I'll just open a new PR when that's ready.

PR #26, opened 2017-04-21
feat: implement PSK message (RFC 16)

andrew-g-za:
First stab at a PSK message implementation.

I'm not convinced the design is correct - opinions gratefully accepted.
It also seems to rely heavily on byte[] - not sure if its worth moving to nio and ByteBuffer?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,13 @@
 machine:
   java:
     version: oraclejdk8
+  post:
+    # apply the JCE unlimited strength policy to allow the PSK 256 bit key length
+    # solution from http://qiita.com/yoskhdia/items/f4702a3abc4467de69b0
+    - curl -L --cookie ''oraclelicense=accept-securebackup-cookie;'  http://download.oracle.com/otn-pub/java/jce/8/jce_policy-8.zip -o /tmp/jce_policy.zip
+    - unzip -o /tmp/jce_policy.zip -d /tmp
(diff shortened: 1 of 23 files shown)


adrianhopebailie:
> Seems we might want to hold this PR until IL-RFC #197 is resolved (Should public headers be removed from PSK)?

Since we have no PSK functionality at all I'd suggest we merge and then adjust when [IL-RFC #197](https://github.com/interledger/rfcs/issues/197) is resolved.

sappenin:
Sounds good to me -- looks like a few test failures, but once that passes I'm good if this gets merged.

andrew-g-za:
Cool, thanks for the feedback.
My bad, I didn't spot the test-case failure the first time :(
I'm pretty sure its related to the 256-bit key length, which requires the Java Cryptography Extension (JCE) Unlimited Strength Jurisdiction Policy Files. I'll try figure out how to get that into CircleCI

andrew-g-za:
JCE policy now deploying in circleci, looks like the tests are passing. I see the previous commit for the OER encoding upset checkstyle, will fix in another submission.

sappenin:
One other thought I had - what if we bump the version to 0.3.0-SNAPSHOT for this commit?

andrew-g-za:
Yup, great idea. Let me get the OER fixes in underneath first, then we can use the PSK change to bump the release to v0.3 as you suggest (I wouldn't want to push this release as v0.3 then immediately have to commit small style changes into it).

sappenin:
+1.  Looks like #27 is good to go.

PR #39, opened 2017-05-12
feat: add code coverage

andrew-g-za:
Adding code coverage support to java-ilp-core. If everyone is happy, i'll do the same for the remaining repos.
Also fixed the checkstyle issue - checkstyle problems will now correctly fail the build on circle-ci.


adrianhopebailie:
@andrew-g-za you can merge, we should fix the failing CheckStyle checks separately

Issue #30, opened 2017-05-10
PSK: Remove withNonce from PskMessageBuilder?


sappenin:
Wondering if we should remove `#withNonce()` and `#withNonce(byte[] nonce)` from `PskMessageBuilder`, and just let the builder implementation always do the right thing with respect to the nonce? (e.g., make `#withNonce()` private and just do that by default).

The nonce is supposed to be cryptographically random, and so it seems a bit risky to allow a caller to send any ole data into there.  Wouldn't it be better to merely have something like this as the builder contract...

```
// Notice no "withNonce" call...
final PskMessage pskMessage = new PskMessageBuilder()
      .addPrivateHeader(privateHeader1)
      .addPublicHeader(publicHeader1)
      .setApplicationData(applicationData)
      .toMessage();
```

... and let the PskMessageBuilder always create random nonce data using `SecureRandom`?  That way the implementation can control the quality of the nonce in all cases.  

Thoughts?

earizon:
+1

andrew-g-za:
The thinking was that we deliberately provided a default implementation - withNonce() - that would do the right thing, but also provide a withNonce(byte[]) for users who wanted to do something special  - for example if a user trusts the bouncy-castle random generator more, or have some custom nonce scheme, or some other special requirement. Having the two methods give it more flexibility, but at the risk that its either abused or simply omitted.

Since the builder is meant to be a convenience, perhaps you are right and it should be more opinionated - it should do the right thing, and if someone wants something special they either implement their own builder or construct PskMessages directly (or wrestle with the headers after using the builder). That way at least there is no risk of the mandatory nonce header being accidentally omitted.

+1 Go for it.

adrianhopebailie:
I think using `SecureRandom` is the right way to go. This way you stick with Java's intentional separation of concerns between the application code and the deployment (and thereby choice of randomness).

The JVM can be configured outside of the application to use different sources of randomness under the hood of `SecureRandom`.

Useful reference: https://tersesystems.com/2015/12/17/the-right-way-to-use-securerandom/

andrew-g-za:
Cool, looks like we are all agreed. Will make the change quickly.

andrew-g-za:
@sappenin - I see there are now two sets of PSK tests, one in org.interledger.ilp.psk and one in org.interledger.psk. I presume i should remove the original set in org.interledger.ilp.psk?

sappenin:
[...]

PR #40, opened 2017-05-12
refactor: remove withNonce from PskMessageBuilder

andrew-g-za:
Changes as per #30 
Note that although the ```.withNonce``` methods have been removed, users can still explicitly add a nonce header via the builder. Maybe that needs to be prevented too?

Ps. note sure whats gotten into git, the PR seems to reference all the changes I've made today, it should really only show 2d5c34e?


sappenin:
Overall, Looks good.  I'd like to prevent users from manually adding headers with the name `Encryption`, `None` or `Key`, although that makes testing a bit more complicated.

I think for testing purposes, we could create a test-only extension of `PskMessageBuilder` and overwrite the `addPublicHeader` methods to not validate against the header-names above.  We would only use that when trying to create a Message for assertion purposes, but otherwise the tests would use the standard builders, which _would_ enforce the header name restrictions.  Thougths?

adrianhopebailie:
> Maybe that needs to be prevented too?

I agree but given the discussion around PSK is still open let's not waste cycles on that. I suggest merging and logging an issue to address this which we might close if the PSK design changes.

Issue #33, opened 2017-05-10
Decide where to put usage/readme info


sappenin:
We need to decide where to put README and otherwise useful information for developer consmption.  

One nice example of a Github wiki is the [Netflix Hystrix](https://github.com/Netflix/Hystrix/wiki) project, where they have light README files and then put most information in the github wiki.  Specifically, [this wiki page](https://github.com/Netflix/Hystrix/wiki/How-To-Use) is very useful when trying to use the library in implementations.

sappenin:
On the previous ILP Java call, we discussed putting documentation for the project into the /docs folder of the project, which would allow publishing to Github pages.  See the [scripts](https://github.com/interledger/rfcs/tree/master/scripts) folder in the RFCs project.

adrianhopebailie:
> Overall, unlike the RFCs project, it's not clear to me what this buys us over just writing our README/Examples/Usages documentation in the Github wiki. The wiki is editable directly inside of the github website (in a browser), and is also versioned. I suppose the only thing missing would be CSS styling? But the downside is reading the doco for this project would take the user away from the actual Github project, which is a slight disadvantage IMHO.

The other advantage is control over edits. Edits would have to be made by PRs. The markdown files would also be published to the web, just need to test how that would work.

sappenin:
> The markdown files would also be published to the web, just need to test how that would work.

Sounds good - where do we store the markdown files?  Maybe in `/docsource`?

sappenin:
I added a wiki-page (for now) to outline the Codec Framework because it wasn't clear where we should put the markdown for the Github Pages stuff.  Once that's worked out, however, we can transition away from the wiki (and turn it off).  

See here: https://github.com/interledger/java-ilp-core/wiki/InterledgerPacket-Handling-Framework

sappenin:
Closing this for now - I think once https://github.com/interledger/java-crypto-conditions/issues/48 is resolved for Javadoc, we'll have a better idea if it makes sense to push usage/README into a Github Pages site for this project (java-ilp-core) and that one (java-crypto-conditions).

PR #44, opened 2017-05-22
Fix #35 and #36

### LINKED ISSUE #35: Fix/Change the structure for Wire/Codec
Per the Java ILP Call and PR #28, we agreed that the `wire` package will be renamed to codec, and the current `codec` package will be pulled up one level so there's only `org.interledger.codec` with everything currently underneath.

Basically, this is just a package-rename task.

### LINKED ISSUE #36: Drop the InterledgerPacket interface
Per the Java ILP call and PR #28, we are going to drop the InterledgerPacket interface so there's a clear separation between the Java domain model and any encoding structures.  Essentially, the codec should be the one to know if a "type" bit should be present or not for codec purposes.

sappenin:
Fixes #35 and #36.  This PR is mostly just moving code around.  The substantive changes are found in InterledgerPacket.java and (its unit tests) and CodecContext.java, which takes advantage of the InterledgerPacket handling framework.

Specifically:

* Fix #35: Remove the “wire” package and move all codec-related interfaces/clases into the “org.interledger.codecs” package.
* Fix #36: Move “InterledgerPacket” to a top-level interledger namespace.  Move “Packet Handler” mechanisms to be internally-defined in InterledgerPacket for clarity.
* Add unit tests for InterledgerPacket.Handler abstract utility classes.

Misc
* Add package-info.java classes where appropriate.
* Formatting, package renames.
* Formatting to appease Checkstyle.


src/main/java/org/interledger/InterledgerPacket.java
```diff
@@ -0,0 +1,245 @@
+package org.interledger;
```
adrianhopebailie: I like this idea. Some examples will help a lot to eek out any design tweaks. Perhaps we should update java-ilp-client to use this as a first pass for some implementation experience?

sappenin: No objections there - I'm going to add a "usage" page on the wikie (for now) and once we get #33 worked out, we can disable the wiki.

PR #49, opened 2017-06-23
Remove dependency on crypto-conditions and fix IPR

adrianhopebailie:
- Removed dependency on crypto-conditions

This was a tough decision but they are adding complexity and actually no longer compatible with some of the binary types (such as IPR) which encode conditions as 32 byte octet strings.

- Fixed IPR

Updated the IPR class to match RFC 11 and added codecs for it

### DIFF
--- a/src/main/java/org/interledger/Condition.java
+++ b/src/main/java/org/interledger/Condition.java
@@ -1,4 +1,4 @@
-package org.interledger.ilp;
+package org.interledger;
 
 import java.util.Arrays;
 
--- a/src/main/java/org/interledger/Condition.java
+++ b/src/main/java/org/interledger/Condition.java
@@ -9,8 +9,8 @@ import java.util.Objects;
  * <p>Interledger relies on conditional payments where each transfer that is part of a payment is
  * conditional upon the presentation of a fulfillment.
  * 
- * <p>The standard for conditions is to use the SHA-256 hash of a pre-image. The pre-image is therefor
- * the fulfillment of the condition.
+ * <p>The standard for conditions is to use the SHA-256 hash of a pre-image. The pre-image is
+ * therefor the fulfillment of the condition.
  * 
  * @see Fulfillment
  */
--- a/src/main/java/org/interledger/Condition.java
+++ b/src/main/java/org/interledger/Condition.java
@@ -1,27 +1,36 @@
 package org.interledger;
 
 import java.util.Arrays;
+import java.util.Objects;
 
 /**
  * The execution condition attached to all transfers in an Interledger payment.
+ * 
+ * <p>Interledger relies on conditional payments where each transfer that is part of a payment is
+ * conditional upon the presentation of a fulfillment.
+ * 
+ * <p>The standard for conditions is to use the SHA-256 hash of a pre-image. The pre-image is therefor
+ * the fulfillment of the condition.
+ * 
+ * @see Fulfillment
  */
 public class Condition {
 
-  private byte[] hash;
+  private final byte[] hash;
 
   /**
    * Create a new {@code Condition} from a SHA-256 hash.
    * 
-   * @param hash The SHA-256 hash of a preimage that is used as the fulfillment of this condition
+   * @param hash The SHA-256 hash of a pre-image that is used as the fulfillment of this condition
    */
   public Condition(byte[] hash) {
 
-    if (hash == null || hash.length != 32) {
+    Objects.requireNonNull(hash, "Hash must not be null!");
+    if (hash.length != 32) {
       throw new IllegalArgumentException("Hash must be 32 bytes.");
     }
 
-    this.hash = new byte[32];
-    System.arraycopy(hash, 0, this.hash, 0, 32);
+    this.hash = Arrays.copyOf(hash, 32);
   }
 
   @Override
@@ -48,7 +57,7 @@ public class Condition {
     if (!Arrays.equals(hash, other.hash)) {
       return false;
     }
-    
+
     return true;
   }
 
@@ -58,9 +67,7 @@ public class Condition {
    * @return a {@code byte[]} of exactly 32 bytes
    */
   public byte[] getHash() {
-    byte[] hash = new byte[32];
-    System.arraycopy(this.hash, 0, hash, 0, 32);
-    return hash;
+    return Arrays.copyOf(this.hash, 32);
   }
 
 }
--- a/src/main/java/org/interledger/Fulfillment.java
+++ b/src/main/java/org/interledger/Fulfillment.java
@@ -3,14 +3,20 @@ package org.interledger;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
+import java.util.Objects;
 
 /**
  * The fulfillment of a {@link Condition}.
+ * 
+ * <p>The standard for Interledger payments is for the fulfillment to be the pre-image of a SHA-256
+ * hash (the condition).
+ * 
+ * <p>The fulfillment (pre-image) must be exactly 32 bytes.
  */
 public class Fulfillment {
 
-  private byte[] preimage;
-  private Condition condition;
+  private final byte[] preimage;
+  private final Condition condition;
 
   /**
    * Create a new Fulfillment from the 32 byte pre-image that is hashed to get the condition.
@@ -19,12 +25,12 @@ public class Fulfillment {
    */
   public Fulfillment(byte[] preimage) {
 
-    if (preimage == null || preimage.length != 32) {
-      throw new IllegalArgumentException("Pre-image must be at least 32 bytes.");
+    Objects.requireNonNull(preimage, "Pre-image must not be null!");
+    if (preimage.length != 32) {
+      throw new IllegalArgumentException("Pre-image must be 32 bytes.");
     }
 
-    this.preimage = new byte[32];
-    System.arraycopy(preimage, 0, this.preimage, 0, 32);
+    this.preimage = Arrays.copyOf(preimage, 32);
 
     try {
       MessageDigest digest = MessageDigest.getInstance("SHA-256");
@@ -42,9 +48,7 @@ public class Fulfillment {
    * @return 32 byte octet string
    */
   public byte[] getPreimage() {
-    byte[] preimage = new byte[this.preimage.length];
-    System.arraycopy(this.preimage, 0, preimage, 0, this.preimage.length);
-    return preimage;
+    return Arrays.copyOf(this.preimage, 32);
   }
 
   /**
@@ -87,7 +91,7 @@ public class Fulfillment {
     if (getClass() != obj.getClass()) {
       return false;
     }
-    
+
     Fulfillment other = (Fulfillment) obj;
     if (condition == null) {
       if (other.condition != null) {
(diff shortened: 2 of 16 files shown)


sappenin:
Wow, initially I didn't love this change, but it does seem to really simplify things, especially if ILP continues to just use Hashlocks. 

I have a few minor nits around javadoc and such, but otherwise this looks good.

I suppose in the future (like when we get java-crypto-conditions published to maven and possibly cleaned-up a bit) we might reconsider depending on java-crypto-conditions again?

src/main/java/org/interledger/Condition.java
```diff
@@ -0,0 +1,66 @@
+package org.interledger;
+
+import java.util.Arrays;
+
+/**
+ * The execution condition attached to all transfers in an Interledger payment.
```
sappenin: Might be worth adding a link to the [crypto-conditions spec](https://datatracker.ietf.org/doc/draft-thomas-crypto-conditions/), and even possibly copying some of the Preimage-sha-256 documentation into here so that it's more clear what a "Condition" is?

src/main/java/org/interledger/Condition.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+public class Condition {
+
+  private byte[] hash;
```
sappenin: Let's make the properties `final` so that these objects are completely thread-safe.

src/main/java/org/interledger/Condition.java
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Create a new {@code Condition} from a SHA-256 hash.
+   * 
+   * @param hash The SHA-256 hash of a preimage that is used as the fulfillment of this condition
+   */
+  public Condition(byte[] hash) {
```
sappenin: What about naming this `PreimageSha256Condition` so that it's clearer that this is something related to the crypto-conditions spec?

src/main/java/org/interledger/Condition.java
```diff
@@ -0,0 +1,66 @@
...
+
+    if (hash == null || hash.length != 32) {
+      throw new IllegalArgumentException("Hash must be 32 bytes.");
+    }
+
+    this.hash = new byte[32];
+    System.arraycopy(hash, 0, this.hash, 0, 32);
```
sappenin: We should prefer the following style for conciseness and portability:

`this.hash = Arrays.copyOf(hash, hash.length);`

src/main/java/org/interledger/Condition.java
```diff
@@ -0,0 +1,66 @@
...
+  private byte[] hash;
+
+  /**
+   * Create a new {@code Condition} from a SHA-256 hash.
+   * 
+   * @param hash The SHA-256 hash of a preimage that is used as the fulfillment of this condition
+   */
+  public Condition(byte[] hash) {
+
+    if (hash == null || hash.length != 32) {
```
sappenin: Can we split this up into two different exceptions, for clarity:

```
Objects.requireNonNull(hash, "hash must not be null!");
if(hash.length != 32) {
   throw new IllegalArgumentException("Hash must be 32 bytes.");
}
```

src/main/java/org/interledger/Condition.java
sappenin: Counter-point: Since we're only using a single type of Condition, the name `Condition` is pretty clean and simple. I'm ok with either way.

src/main/java/org/interledger/Condition.java
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  /**
+   * Get the SHA-256 hash of this condition.
+   * 
+   * @return a {@code byte[]} of exactly 32 bytes
+   */
+  public byte[] getHash() {
+    byte[] hash = new byte[32];
+    System.arraycopy(this.hash, 0, hash, 0, 32);
```
sappenin: This should use the style suggested above:

```
return Arrays.copyOf(hash, hash.length);
```

src/main/java/org/interledger/Fulfillment.java
```diff
@@ -0,0 +1,106 @@
...
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
+
+/**
+ * The fulfillment of a {@link Condition}.
+ */
+public class Fulfillment {
+
+  private byte[] preimage;
```
sappenin: These two properties should be final.

src/main/java/org/interledger/Fulfillment.java
```diff
@@ -0,0 +1,106 @@
...
+  private Condition condition;
+
+  /**
+   * Create a new Fulfillment from the 32 byte pre-image that is hashed to get the condition.
+   * 
+   * @param preimage A 32 byte octet string
+   */
+  public Fulfillment(byte[] preimage) {
+
+    if (preimage == null || preimage.length != 32) {
```
sappenin: Can we separate these two for clarity:

```
Objects.requireNonNull(preimage, "preimage must not be null!");
if(preimage.length != 32) {
   throw new IllegalArgumentException("Preimage must be 32 bytes.");
}
```

src/main/java/org/interledger/Fulfillment.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+public class Fulfillment {
+
+  private byte[] preimage;
+  private Condition condition;
+
+  /**
+   * Create a new Fulfillment from the 32 byte pre-image that is hashed to get the condition.
+   * 
+   * @param preimage A 32 byte octet string
```
sappenin: Might be worth clarifying what a `pre-image` is. Without being familiar with either crypto in general, or the crypto-conditions spec, this term may not make sense.

src/main/java/org/interledger/Fulfillment.java
```diff
@@ -0,0 +1,106 @@
...
+   */
+  public Fulfillment(byte[] preimage) {
+
+    if (preimage == null || preimage.length != 32) {
+      throw new IllegalArgumentException("Pre-image must be at least 32 bytes.");
+    }
+
+    this.preimage = new byte[32];
+    System.arraycopy(preimage, 0, this.preimage, 0, 32);
```
sappenin:
```
this.preimage = Arrays.copy(preimage, preimage.length);
```

src/main/java/org/interledger/Fulfillment.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  /**
+   * Get the raw pre-image (safe copy).
+   * 
+   * @return 32 byte octet string
+   */
+  public byte[] getPreimage() {
+    byte[] preimage = new byte[this.preimage.length];
+    System.arraycopy(this.preimage, 0, preimage, 0, this.preimage.length);
```
sappenin: Use other style for Array copying....see above.

src/main/java/org/interledger/codecs/oer/ipr/InterledgerPaymentRequestOerCodec.java
```diff
@@ -0,0 +1,52 @@
...
+import org.interledger.ilp.InterledgerPayment;
+import org.interledger.ipr.InterledgerPaymentRequest;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Objects;
+
+/**
+ * An implementation of {@link Codec} that reads and writes instances of {@link InterledgerAddress}.
```
sappenin: `InterledgerAddress` should be renamed to `InterledgerPaymentRequest` in the javadoc.

src/main/java/org/interledger/codecs/oer/ilp/ConditionOerCodec.java
```diff
@@ -0,0 +1,39 @@
...
+import org.interledger.codecs.ConditionCodec;
+import org.interledger.codecs.oer.OerUint256Codec.OerUint256;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Objects;
+
+/**
+ * An implementation of {@link Codec} that reads and writes instances of {@link InterledgerAddress}.
```
sappenin: `InterledgerAddress` should be renamed to `Condition` in the javadoc. So, should read:

```
An implementation of {@link Codec} that reads and writes instances of {@link Condition}.
```

src/main/java/org/interledger/Condition.java
adrianhopebailie: Will fix

src/main/java/org/interledger/Condition.java
adrianhopebailie: Will fix

src/main/java/org/interledger/Condition.java
adrianhopebailie: Yeh, basically there was a decision to only support SHA-256 conditions so I want to keep this simple. We event restrict the size of fulfillments to 32 bytes.

src/main/java/org/interledger/Condition.java
adrianhopebailie: Will fix

src/main/java/org/interledger/Condition.java
adrianhopebailie: Will fix

src/main/java/org/interledger/Fulfillment.java
adrianhopebailie: Will fix

src/main/java/org/interledger/Fulfillment.java
adrianhopebailie: Will fix

src/main/java/org/interledger/Fulfillment.java
adrianhopebailie: Will do

src/main/java/org/interledger/Fulfillment.java
adrianhopebailie: Will fix

src/main/java/org/interledger/Fulfillment.java
adrianhopebailie: Will fix

src/main/java/org/interledger/codecs/oer/ipr/InterledgerPaymentRequestOerCodec.java
adrianhopebailie: :) Will fix

src/main/java/org/interledger/codecs/oer/ilp/ConditionOerCodec.java
adrianhopebailie: Will fix

src/main/java/org/interledger/codecs/oer/OerUint256Codec.java
```diff
@@ -0,0 +1,119 @@
...
+   * @throws IllegalArgumentException If the input has a value greater than 18446744073709551615.
+   */
+  @Override
+  public OerUint256 read(final CodecContext context, final InputStream inputStream)
+      throws IOException {
+    Objects.requireNonNull(context);
+    Objects.requireNonNull(inputStream);
+
+    final byte[] returnable = new byte[32];
+    inputStream.read(returnable);
```
andrew-g-za: inputstream.read(byte[]) does not guarantee that it will fill the array, so we should check that it read the required number.

adrianhopebailie: Good catch, fixed

PR #50, opened 2017-06-27
Earizon code review new condition

earizon:
I adapted the existing java-vertx-ledger code to the new changes and found some needed changes in java-ilp-core. This PR is the summary of the code review I realized.

NOTE: The static utility method Condition.parse(URI...) is needed by the ledger when receiving the condition through the REST API interfaces:
  - PUT /.../transfers/:transferUUID
    body { .... , "execution_condition": "ni:...." }
  - get /.../transfers/byExecutionCondition/..../

### DIFF
--- a/src/main/java/org/interledger/Condition.java
+++ b/src/main/java/org/interledger/Condition.java
@@ -1,7 +1,13 @@
 package org.interledger;
 
+import java.net.URI;
 import java.util.Arrays;
+import java.util.Base64;
 import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.interledger.cryptoconditions.uri.URIEncodingException;
 
 /**
  * The execution condition attached to all transfers in an Interledger payment.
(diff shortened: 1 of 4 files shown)


adrianhopebailie:
@earizon thanks for this. Implementation experience is very helpful to validate the changes.

I am looking at the JS code and I think that the conversion from `ni:` URI is actually not part of the standard but part of the 5 bells implementation. I think that code should be part of your adapter not the core lib.

Where is your ledger code? I am keen to test against it a bit

earizon:
@adrianhopebailie The code is @ https://github.com/earizon/java-vertx-ledger/

The README.md contains all the info to setup & run the ledger (or so I hope)

the "dev_docs" directory has some interesting notes about the code. Specially the architecture notes:
https://github.com/earizon/java-vertx-ledger/blob/master/dev_docs/architecture_notes.txt

adrianhopebailie:
@earizon do you agree that this conversion should be in the ledger adaptor code?

earizon:
@adrianhopebailie Yes, I just put it moving from a helper function in crypto-condition. No problem on my side.

PR #51, opened 2017-06-27
Add PSK functions for deriving keys

adrianhopebailie:
Add PSK functions for deriving shared key
Add IPR builder


src/main/java/org/interledger/psk/PskParams.java
```diff
@@ -0,0 +1,28 @@
+package org.interledger.psk;
+
+/**
+ * The parameters derived from a receiver secret for use in the PSK protocol.
+ */
```
sappenin: Would be nice to have a link to the PSK Protocol.

For example:

```
/**
* The parameters derived from a receiver secret for use in the PSK protocol.
*
* @see "https://github.com/interledger/rfcs/tree/master/0016-pre-shared-key"
*/
```

src/main/java/org/interledger/psk/io/PskUtils.java
```diff
@@ -1,66 +1,197 @@
...
 import javax.crypto.Cipher;
+import javax.crypto.Mac;
 import javax.crypto.SecretKey;
 import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
 
 /**
  * Convenience methods for PSK messages.
  */
 public class PskUtils {
```
sappenin: We should have unit tests for this class (unless I'm missing that)

src/main/java/org/interledger/psk/io/PskUtils.java
```diff
@@ -1,66 +1,197 @@
...
+  /* HMAC SHA-256 spec */
+  public static final String HMAC_ALGORITHM = "HmacSHA256";
+
+  /**
+   * Generate PSK parameters from receiver secret.
+   * 
+   * @param receiverSecret The local secret used by the receiver
+   * @return PSK parameters derived from secret
+   */
+  public static PskParams getPskParams(byte[] receiverSecret) {
```
sappenin: Minor nit: we should add `Objects.requireNonNull` as a precondition check.

src/main/java/org/interledger/ipr/InterledgerPaymentRequestBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+import java.time.ZonedDateTime;
+import java.util.Objects;
+
+public class InterledgerPaymentRequestBuilder {
+
+  private final PskParams pskParams;
+  private final PskMessageBuilder messageBuilder;
+
+  // ILP Packet headers
+  private final InterledgerPayment.Builder packetBuilder;
```
adrianhopebailie: I agree. Will fix.

src/main/java/org/interledger/ipr/InterledgerPaymentRequestBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+   * receiverSecret.
+   * 
+   * @param receiverAddressPrefix ILP Address of the receiver (will have a unique id and token
+   *        appended to allow the fulfillment to be generated when the packet is received later).
+   * @param receiveAmount The amount to receive.
+   * @param expiry The expiry of the payment request (used to populate the Expires-At header in the
+   *        transport layer envelope).
+   * @param receiverSecret The receiver secret as defined in the PSK specification.
+   */
+  public InterledgerPaymentRequestBuilder(InterledgerAddress receiverAddressPrefix,
```
adrianhopebailie: All params are required. Will fix the rest.

src/main/java/org/interledger/ipr/InterledgerPaymentRequestBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+   * 
+   * <p>Calling this will result in the internal PSK message being built (and encrypted unless
+   * encryption is disabled).
+   * 
+   * <p>After the PSK message is built the ILP Packet is built and OER encoded before the Condition
+   * is generated.
+   * 
+   * @return an Interledger Payment Request.
+   */
+  public InterledgerPaymentRequest getIpr() {
```
adrianhopebailie: Will fix

src/main/java/org/interledger/ipr/InterledgerPaymentRequestBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+
+    PskMessageWriter writer =
+        (this.encrypted) ? PskWriterFactory.getEncryptedWriter(this.pskParams.getSharedKey())
+            : PskWriterFactory.getUnencryptedWriter();
+
+    packetBuilder.data(writer.writeMessage(messageBuilder.toMessage()));
+    InterledgerPayment packet = packetBuilder.build();
+
+    // TODO: It's a pity we have to encode the packet to get the condition here
```
adrianhopebailie: The packet has to be encoded to get the condition. Need to think about a better architecture here, perhaps an alternative version of the condition that encodes the packet and generates the hashes late but probabaly overkill for now

src/test/java/org/interledger/psk/AesEncryptionTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Test that locally installed provider is working as expected.
+ * 
+ * <p>To use 256-bit AES keys you must have Java Cryptography Extension (JCE) Unlimited Strength
+ * Jurisdiction Policy Files installed.
+ * 
+ * <p>{@link http://www.oracle.com/technetwork/java/javase/downloads/jce8-download-2133166.html}
+ */
+public class AesEncryptionTest {
```
andrew-g-za: maybe im missing something, but is this test **just** to act as a canary for having JCE installed? I guess its a nice convenience for people setting up, since getting NoSuchAlgorithm can be misleading.
Perhaps lets rename the test to be more meaningful, like Aes256KeyLengthTest, and maybe have an explicit check
try {
... do something that requires AES-256
} catch(NoSuchAlgorithmException nsa) {
fail("AES-256 key could not be used, please check JCE policy is installed, nsa);
}

adrianhopebailie: > is this test just to act as a canary for having JCE installed
Yes, some good suggestions, will fix this up.

src/main/java/org/interledger/ipr/InterledgerPaymentRequestBuilder.java
sappenin: Sounds good.

PR #52, opened 2017-06-30
feat: add OER support for ILQP packets

andrew-g-za:
@sappenin - please take a look at OerIA5StringCodec and OerLengthPrefixCodec, i made a few changes to ensure that length indicators are checked more carefully, and (i think) improved performance for the string codec (encode to byte[] once instead of twice).

I have an[ issue open for the GeneralizedTime](https://github.com/interledger/rfcs/issues/230) format referenced in the LiquidityResponse, i figure lets get the rest in so long while that gets resolved.


src/main/java/org/interledger/codecs/oer/OerIA5StringCodec.java
```diff
@@ -33,7 +33,9 @@ public OerIA5String read(
 
     // Detect the length of the encoded IA5String, and move the buffer index to the correct spot.
     final int length = context.read(OerLengthPrefix.class, inputStream).getLength();
-    final String result = this.toString(inputStream, length);
+    
+    /* beware the 0-length string */
+    final String result = (length == 0 ? "" : this.toString(inputStream, length));
```
sappenin: Nice catch!

src/main/java/org/interledger/codecs/oer/OerUint64Codec.java
```diff
@@ -33,12 +35,14 @@ public OerUint64 read(final CodecContext context, final InputStream inputStream)
...
 
-    long value = 0;
-    for (int i = 0; i < 8; i++) {
-      value <<= Byte.SIZE;
-      value |= (inputStream.read() & 0xFF);
+    byte[] value = new byte[8];
+    int read = inputStream.read(value);
+    
+    if (read != 8) {
+      throw new IOException("unexpected end of stream. expected 8 bytes, read " + read);
```
sappenin: I think lines 38-43 can be replaced with a single line to achieve the same effect:

```
inputStream.read(value, 0, 8);
```

sappenin: Oops, my bad - I was wrong. But this might be preferable to help the reader not exceed 8 bytes:

```
int read = inputStream.read(value, 0, 8);
if (read != 8) {
    throw new IOException("unexpected end of stream. expected 8 bytes, read " + read);
}

```

sappenin: Actually, on second though, maybe we want to just consume whatever is in the InputStream to account for all the edge cases. You can disregard this comment.

PR #54, opened 2017-07-14
PSK refactor

adrianhopebailie:
This PR has a some extensive changes but mostly it is a moving things around for consistency.

1. The builders and core interfaces all follow the same pattern now
2. The PSK functions are better abstracted
 - Encryption and decryption are not coupled with encoding and decoding
 - Codecs implemented using the codec classes
 - Keys/secrets held in a context object that performs PSK functions
 - PSK Message objects are encrypted/decrypted as objects

@sappenin @andrew-g-za  Please start reviewing while I work on test vectors and checkstyle.

For a good example of everything tying together see the test in:
`org.interledger.ipr.InterledgerPaymentRequestEndToEndTest`

### Tests are failing
Some unit tests are failing due to enforcing 32 byte size on the receiver secret and shared secret. The known good values from the JS impl are no longer usable as they were the wrong size. The tests are correct but the vectors need to be updated.

### Checkstyle
Not yet up to date on the test code but the main code is all passing


build.gradle
```diff
@@ -1,6 +1,6 @@
 group = 'org.interledger'
 description = 'Interledger Protocol Core'
-version = '0.5.0-SNAPSHOT'
+version = '0.5.1-SNAPSHOT'
```
sappenin: Not sure when this PR will be fully-baked, but it's likely that #55 will be merged before this one, so something to keep in mind - you'll need to pull-in those changes to this build file.

src/test/java/org/interledger/FulfillmentTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.junit.Test;
+
+import java.util.Base64;
+
+public class FulfillmentTest {
+
+  private static byte[] TEST_PREIMAGE = new byte[32];
+  private static Condition TEST_CONDITION = Condition.builder()
+      .hash(Base64.getUrlDecoder().decode("Zmh6rfhivXdsj8GLjp-OIAiXFIVu4jOzkCpZHQ1fKSU")).build();
```
sappenin: It might be preferable to add a 2nd additional test that generates 256 bits of random byte data here for the preimage so that each time the test runs, it validates a different preimage? Then we would have some static coverage (what's there now) as well as on-going assurances across different hash values.

src/test/java/org/interledger/codecs/CodecContextFactoryTest.java
```diff
@@ -36,18 +38,21 @@ public void interledger() throws Exception {
     assertTrue(context.hasRegisteredCodec(InterledgerAddress.class));
     assertTrue(context.hasRegisteredCodec(InterledgerPacketType.class));
     assertTrue(context.hasRegisteredCodec(InterledgerPayment.class));
+    assertTrue(context.hasRegisteredCodec(InterledgerPaymentRequest.class));
 
     // TODO: Add ILQP!
-    //    assertTrue(context.readable(InterledgerPacketType.ILQP_QUOTE_LIQUIDITY_REQUEST_TYPE));
+    // assertTrue(context.readable(InterledgerPacketType.ILQP_QUOTE_LIQUIDITY_REQUEST_TYPE));
```
sappenin: Looks like these assertions didn't get enabled when the ILQP code was merged - we should uncomment these lines.

src/main/java/org/interledger/ilp/InterledgerPayment.java
```diff
@@ -64,21 +58,29 @@
...
+
+  /**
    * A builder for instances of {@link InterledgerPayment}.
    */
   class Builder {
 
-    private InterledgerAddress destinationAccount;
-    private Long destinationAmount;
-    private byte[] data;
+    protected InterledgerAddress destinationAccount;
```
adrianhopebailie: Old code had sub-classes. Fixed

README.md
```diff
@@ -15,6 +15,118 @@ For more information about Interledger specifications that underpin this library
 ## Usage
 For more detail about how to use this library, consult the [wiki](https://github.com/interledger/java-ilp-core/wiki)
 
+The library offers interfaces for all core objects and default builders that produce immutable instances of these objects.
```
sappenin: This is awesome!!

README.md
```diff
@@ -15,6 +15,118 @@ For more information about Interledger specifications that underpin this library
...
+//Check if an address starts with a specific prefix
+if( address.startsWith("test") ) {
+  //Testnet address...
+}
+
+```
+
+### Conditions and Fulfillments
+
+Conditions can be derived from a fullfillment or directly from the underlying hash
```
sappenin: We should put a link to the crypto-conditions project and to the PSK rfc so that people can reference what these things are.

adrianhopebailie: We aren't actually using crypto-conditions anymore. Have added a link to the RFC though.

adrianhopebailie:
This is becoming unwieldy given the number of commits. Replaced with a single squashed commit and PR https://github.com/interledger/java-ilp-core/pull/58/

PR #58, opened 2017-07-24
Refactor for consistency and code style

adrianhopebailie:
- Used auto-formatter based on Check-style rules
 - The builders and core interfaces all follow the same pattern now
 - The PSK functions are better abstracted
   - Encryption and decryption are not coupled with encoding and decoding
   - Codecs implemented using the codec classes
   - Keys/secrets held in a context object that performs PSK functions
   - PSK Message objects are encrypted/decrypted as objects
- Added examples to README

This replaces PR 54 which has become quite unwieldy. All commits squashed into one now and rebased onto `development`


src/main/java/org/interledger/psk/PskMessage.java
```diff
@@ -0,0 +1,410 @@
...
+   *
+   * @return the <b>public</b> headers whose name matches the parameter, or an empty list.
+   */
+  List<Header> getPublicHeaders(String headerName);
+
+  /**
+   * Returns a list of all <b>private</b> headers in the message, provided that either the message
+   * was <b>not</b> encrypted, or the private portion of the message has been decrypted by the
+   * receiver.
+   */
```
sappenin: Minor nit, but we should duplicate the Javadoc clarification from `getPrivateHeaders(headerName)` to overtly state the returned list is empty in certain circumstances:

```
 @return the <b>private</b> headers for this PskMessage, or an empty list.
```

src/main/java/org/interledger/Fulfillment.java
```diff
@@ -7,104 +7,152 @@
 
 /**
  * The fulfillment of a {@link Condition}.
```
sappenin: Adding my (edited) Javadoc suggestion from the previous PR:

```
/**
     * A Java implementation of a Crypto-Conditions Fulfillment, which enables the concept of a hashlock (see link below). 
     * 
     * In Hashlock terms, a fulfillment is the "key" which, when published, allows other primitives (in this case a
     * {@link Condition}) to become "opened" or fulfilled, making it possible for various operations to be
     * performed in response to the publishing of a fulfillment.
     * 
     * The standard for Interledger payments is for a fulfillment to be the pre-image of a SHA-256
     * hash (the condition).
     * 
     * Note that Fulfillments can always generate their associated Condition (though not vice versa).
     *
     * @see "https://en.bitcoin.it/wiki/Hashlock"
     * @see "https://tools.ietf.org/html/draft-thomas-crypto-conditions-03"
     */
```

sappenin: I think it's helpful to describe what a `Fulfillment` and `Condition` actually are, per something like the above. Before, when we relied upon crypto-conditions project, people could look there and try to understand that stuff. Now that we don't have those links, it can be extra confusing.

I've noticed two impediments to outsider's understanding of these concepts. First, Hashlocks are a commonly understood concept if you're familiar with cryptocurrencies, but for those who aren't, the concept is foreign. Next, mapping that concept to the CryptoCondition spec is another leap.

So, simply talking about a "Condition" and and "Fulfillment" is often like speaking a foreign language, especially around people who are new to Interledger.

My $0.02.

src/main/java/org/interledger/Condition.java
```diff
@@ -5,69 +5,103 @@
...
+      private final byte[] hash;
+
+      public Impl(Builder builder) {
+
+        Objects.requireNonNull(builder.hash, "Hash must not be null!");
+        if (builder.hash.length != 32) {
+          throw new IllegalArgumentException("Hash must be 32 bytes.");
+        }
+
+        this.hash = Arrays.copyOf(builder.hash, 32);
```
andrew-g-za: Im just wondering - if the builder made a copy of the hash, is it necessary to make a copy from the builder? Its certainly the safest, but all that array copying does add up performance wise.

src/main/java/org/interledger/Condition.java
```diff
@@ -5,69 +5,103 @@
...
-public class Condition {
+public interface Condition {
 
-  private final byte[] hash;
+  /**
+   * Get the default builder.
+   *
+   * @return a {@link Builder} instance.
+   */
+  static Builder builder() {
```
andrew-g-za: silly question, but wont this make the builder package private?

src/main/java/org/interledger/Condition.java
```diff
@@ -5,69 +5,103 @@
...
-   * 
-   * @param hash The SHA-256 hash of a pre-image that is used as the fulfillment of this condition
+   * Build a new Condition using the provided hash
+   *
+   * @param hash A SHA-256 hash representing a Condition.
+   *
+   * @return a {@link Condition} instance.
    */
-  public Condition(byte[] hash) {
+  static Condition of(byte[] hash) {
```
andrew-g-za: ditto - wouldnt this be package-private?

src/main/java/org/interledger/Fulfillment.java
```diff
@@ -7,104 +7,152 @@
...
-      throw new IllegalArgumentException("Pre-image must be 32 bytes.");
-    }
-
-    this.preimage = Arrays.copyOf(preimage, 32);
-
-    try {
-      MessageDigest digest = MessageDigest.getInstance("SHA-256");
-      byte[] hash = digest.digest(preimage);
-      condition = new Condition(hash);
+  static Builder builder() {
```
andrew-g-za: isnt this be package-private?

src/main/java/org/interledger/Fulfillment.java
```diff
@@ -7,104 +7,152 @@
...
-      throw new RuntimeException(e);
-    }
+  /**
+   * Build a new Fulfillment using the provided preimage.
+   *
+   * @param preimage The preimage representing the fulfillment
+   *
+   * @return a  {@link Fulfillment} instance
+   */
+  static Fulfillment of(final byte[] preimage) {
```
andrew-g-za: ditto - isnt this package-private?

src/main/java/org/interledger/InterledgerAddress.java
```diff
@@ -29,7 +30,27 @@
 public interface InterledgerAddress {
 
   /**
-   * Return this address's value as a non-null {@link String}.  For example:
+   * Constructor to allow quick create of String.
+   *
+   * @param value String representation of an Interledger Address
+   */
+  static InterledgerAddress of(final String value) {
```
andrew-g-za: more questions on package-private

src/main/java/org/interledger/InterledgerAddress.java
```diff
@@ -29,7 +30,27 @@
...
+        .value(value)
+        .build();
+  }
+
+  /**
+   * Get the default builder.
+   *
+   * @return a {@link Builder} instance.
+   */
+  static Builder builder() {
```
andrew-g-za: last one- you get the picture :)

src/main/java/org/interledger/InterledgerAddress.java
```diff
@@ -109,12 +133,151 @@ default boolean startsWith(final InterledgerAddress interledgerAddress) {
...
+    }
+
+    /**
+     * Assign a new value to this builder.
+     *
+     * @param value A {@link String} representing this builder's "value", which is the string
+     *              version of an Interledger Address.
+     */
+    public Builder value(final String value) {
+      this.value = value;
```
andrew-g-za: hmm maybe Objects.requireNonNull ?
Are empty strings valid?

src/main/java/org/interledger/InterledgerAddress.java
```diff
@@ -109,12 +133,151 @@ default boolean startsWith(final InterledgerAddress interledgerAddress) {
...
+      }
+
+      @Override
+      public int hashCode() {
+        return value.hashCode();
+      }
+
+      @Override
+      public String toString() {
+        return "Impl{"
```
andrew-g-za: i just realized that i think all of our concrete implementations .toString have "Impl"  as the prefix, which doesnt really help - we shouldnt we have the interface name as a prefix, e.g. "InterLedgerAddress.Impl"

src/main/java/org/interledger/codecs/AbstractCodec.java
```diff
@@ -19,11 +19,12 @@ public AbstractCodec(final int typeId) {
   }
 
   /**
-   * Read an object from the buffer according to the rules defined in the {@link CodecContext}.
+   * Read an object of the buffer according to the rules defined in the {@link CodecContext}.
```
andrew-g-za: eh, we went from "Read an object from the buffer..." to "Read an object of the buffer..."? or am i not reading the diff correctly?

src/main/java/org/interledger/Fulfillment.java
adrianhopebailie: Gonna leave that off now and we can discuss further and address as part of #61.

I'm worried that talking about crypto-conditions will create more confusion given that we only use hashlocks but agree that we need better documentation on the subject.

src/main/java/org/interledger/codecs/CodecContext.java
```diff
@@ -97,36 +104,68 @@ public InterledgerPacket read(final InputStream inputStream) throws IOException
...
+   * @param <T>  The type of object to return, based upon the supplied type of {@code type}.
+   *
+   * @return An instance of {@link T}.
+   */
+  public <T> T read(final Class<T> type, final byte[] data) {
+    Objects.requireNonNull(type);
+    Objects.requireNonNull(data);
+
+    try (ByteArrayInputStream bais = new ByteArrayInputStream(data)) {
+      if (InterledgerPacket.class.isAssignableFrom(type)) {
```
andrew-g-za: if this check is done in read(final Class<T> type, final InputStream inputStream), do we need it here too? cant we just go 
try(ByteArrayInputStream bais ....) {
 return read(type, bais);
}
?

src/main/java/org/interledger/codecs/oer/OerIA5StringCodec.java
```diff
@@ -41,15 +39,15 @@ public OerIA5String read(
...
+  public void write(final CodecContext context, final OerIA5String instance,
+      final OutputStream outputStream) throws IOException {
 
     Objects.requireNonNull(context);
     Objects.requireNonNull(instance);
     Objects.requireNonNull(outputStream);
 
-    final byte[] data = instance.getValue().getBytes();
+    final byte[] data = instance.getValue()
+        .getBytes();
```
andrew-g-za: it looks like your formatter has gone mad - surely this is not the best way to format this line (and many others too?)
We should probably look into why the formatters arent stable.

src/main/java/org/interledger/codecs/psk/PskMessageBinaryCodec.java
```diff
@@ -0,0 +1,267 @@
...
+        if (byteValue == LF) {
+          break;
+        } else {
+          if (readCarriageReturn) {
+            // Not end of the line but we read a CR previously so write it before the current byte
+            bos.write(CR);
+            readCarriageReturn = false;
+          }
+        }
+        bos.write(byteValue);
```
andrew-g-za: its equivalent because of the continues / breaks in the other if statements, but couldnt we neaten this up to

if (bytevalue == CR) {... continue;}
if (bytevalue == LF) {break;}
if (readCarriageReturn {...} else {bos.write(byteValue);}

we remove one level of nesting, and its clear that either we read a CR and write it back, or we write whatever byte we read.

adrianhopebailie:
@andrew-g-za thanks for the review. Will address in a new PR!

src/main/java/org/interledger/InterledgerAddress.java
sappenin: FYI - the `public` modifier is redundant on an interface because it's implied.

For years while I was an eclipse user, I would add `public` to all interface methods (and other redundant modifiers like `public static final` on a String in an interface). Then, when I switched to IntelliJ, the default setting is to grey-out redundant modifiers, and now I've come to love that style.

Yet another reason you'll come to love IntelliJ!!

src/main/java/org/interledger/Condition.java
adrianhopebailie: No, interfaces are static so everything is public

src/main/java/org/interledger/Condition.java
adrianhopebailie: Ditto :)

src/main/java/org/interledger/Fulfillment.java
adrianhopebailie: As above

src/main/java/org/interledger/Fulfillment.java
adrianhopebailie: Ditto

src/main/java/org/interledger/InterledgerAddress.java
adrianhopebailie: same answer :)

src/main/java/org/interledger/Condition.java
adrianhopebailie: It's possible to call build multiple times on the builder and produce, what should be, immutable copies of the same Condition so I think you need both.

But, that said I should be able to pass the builder an object, call `build()` then change the object I provided to the builder and call `build()` again to get a different outcome. 

That seems a bit "unintentional" but strictly speaking is correct. WDYT? @sappenin @earizon

src/main/java/org/interledger/InterledgerAddress.java
adrianhopebailie: Yes :)

src/main/java/org/interledger/InterledgerAddress.java
adrianhopebailie: I've moved most of the checks into the `build()` function per discussion with @sappenin.

The idea is that the builder should only throw when you call build so you can provide it with objects that may change between creating the builder and calling build.

src/main/java/org/interledger/InterledgerAddress.java
adrianhopebailie: Good idea!

src/main/java/org/interledger/codecs/AbstractCodec.java
adrianhopebailie: Weird, will fix

src/main/java/org/interledger/codecs/CodecContext.java
adrianhopebailie: @sappenin this is actually your code, just looks like an addition due to formatting.
Thoughts?

src/main/java/org/interledger/codecs/oer/OerIA5StringCodec.java
adrianhopebailie: It took me ages to get my formatter to do this! 😄 

This formatting is how we like the builder's to look but it seems it's impossible to have a rule that works for builders but not for other chained method calls.

Maybe need to get the formatter to ignore and leave it to use mortals to do correctly?

src/main/java/org/interledger/Condition.java
andrew-g-za: Yeah i didnt think about calling the builder multiple times - we definitely need the copy then, please keep it.

src/main/java/org/interledger/InterledgerAddress.java
andrew-g-za: Its such a habit for me to do it, i forgot about the implicit public modifier. Here's the official word from the [spec](http://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html)

> Every method declaration in the body of an interface is implicitly public (§6.6). It is permitted, but discouraged as a matter of style, to redundantly specify the public modifier for a method declaration in an interface.

I'll keep that one in mind :)

src/main/java/org/interledger/codecs/oer/OerIA5StringCodec.java
andrew-g-za: I'd rather have a universal standard that just works, even if it does horribly ugly things like that. I wonder if we can include a formatter in the project that the IDE's would use auto-magically?

adrianhopebailie:
Post-merge comments addressed in https://github.com/interledger/java-ilp-core/pull/66

PR #66, opened 2017-07-28
Address comments from @andrew-g-za

adrianhopebailie:


src/main/java/org/interledger/ilp/InterledgerError.java
```diff
@@ -16,15 +16,15 @@
   // This way our code can differentiate between a coding error
   // (developer sent a null by mistake) and the real intention of not
   // sending selfAddress.
-  static final InterledgerAddress TRIGGERING_ILP_NODE =
+  private static final InterledgerAddress TRIGGERING_ILP_NODE =
```
sappenin: I don't really understand this method. Could we just use a `Set` and get rid of the whole `TRIGGERING_ILP_NODE`? I'm not comfortable having a reserved address -- even if we think nobody will use it, it just doesn't seem right.

I see mention of trying to avoid nulls as another justification - in that case, we should have non-null precondition checks so that the code breaks well before production (like in a unit test).

sappenin: I don't really understand this internal address. Could we just use a `Set` for the `forwardedBy` and get rid of the whole `TRIGGERING_ILP_NODE`? 

I'm not comfortable having a reserved Interledger address inside of here -- even if we think nobody will use it, we can't really guarantee that, and since we're trying to use a reserved address, it indicates we're tackling this problem incorrectly.

Additionally, I see mention of trying to avoid nulls as another justification - in that case, we should have non-null precondition checks so that the code breaks well before production (like in a unit test).

adrianhopebailie: That's not my code. 

@earizon can you have a look at @sappenin 's comments?

PR #67, opened 2017-08-02
Improve invalid address logging in InterledgerAddress

sappenin:


adrianhopebailie:
Seem to be some CheckStyle issues though

pom.xml
```diff
@@ -4,7 +4,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.interledger</groupId>
   <artifactId>java-ilp-core</artifactId>
-  <version>0.5.0-SNAPSHOT</version>
+  <version>0.6.1-SNAPSHOT</version>
```
andrew-g-za: im sure we've discussed this before but i seem to have forgotten - are we still checking in POMs at all? If we are using gradle, why bother?

adrianhopebailie: The idea was originally that if you don't use Gradle you check the project out and still build/install it with Maven.

src/main/java/org/interledger/InterledgerAddress.java
```diff
@@ -88,7 +88,7 @@ default boolean startsWith(final String addressSegment) {
    */
   default boolean startsWith(final InterledgerAddress interledgerAddress) {
     Objects.requireNonNull(interledgerAddress, "interledgerAddress must not be null!");
-    return this.startsWith(interledgerAddress.toString());
+    return this.startsWith(interledgerAddress.getValue());
```
sappenin: Just pushed a few more fixes to this PR -- I ran into a bug with this yesterday where the toString() emits JSON, and so this line never returns `true`. I added unit-test coverage for this condition as well.

src/main/java/org/interledger/InterledgerAddress.java
```diff
@@ -202,7 +202,11 @@ private Impl(final Builder builder) {
 
         if (!isValidInterledgerAddress(builder.value)) {
           throw new IllegalArgumentException(
-              "Invalid characters in address.  Reference Interledger RFC-15 for proper format.");
+              String.format(
```
sappenin: Had to change this formatting to placate Checkstyle...

:(

src/main/java/org/interledger/InterledgerAddress.java
```diff
@@ -271,7 +275,7 @@ public int hashCode() {
 
       @Override
       public String toString() {
-        return "InterledgerAddress.Impl{value='" + value + "'}";
+        return this.getValue();
```
sappenin: Any objections to this change? I think I advocated (in the past) for the opposite of what I'm about to say, but for `InterledgerAddress`, simply emitting the `value` as the output of `toString` seems like what I would expect as a developer. Another few arguments in favor of this change:

* Seeing the JSON version of itself doesn't seem terribly helpful in this case since there's only 1 field -- the "value".
* If we emit `value` when calling toString, it's generally pretty obvious that this is an `InterledgerAddress` due to the nature of the format of an InterledgerAddress (so, no need to overtly state that in a JSON payload with one value).

src/main/java/org/interledger/codecs/oer/OerUint32Codec.java
```diff
@@ -48,11 +48,11 @@ public OerUint32 read(final CodecContext context, final InputStream inputStream)
    * upper bound is not greater than 4294967295 and the constraint is not extensible, the integer
    * value is encoded as an unsigned binary integer in four octets.
    *
-   * @param context An instance of {@link CodecContext}.
-   * @param instance An instance of {@link OerUint64}.
+   * @param context      An instance of {@link CodecContext}.
+   * @param instance     An instance of {@link OerUint32}.
```
sappenin: This was breaking Checkstyle...

src/main/java/org/interledger/codecs/oer/ilqp/QuoteLiquidityResponseOerCodec.java
```diff
@@ -25,9 +25,9 @@
 
 /**
  * An implementation of {@link Codec} that reads and writes instances of
- * {@link QuoteLiquidityResponse}. in OER format.
- * 
- * @see https://github.com/interledger/rfcs/blob/master/asn1/InterledgerQuotingProtocol.asn
+ * {@link QuoteLiquidityResponse} to/from ASN.1 OER format.
+ *
+ * @see "https://github.com/interledger/rfcs/blob/master/asn1/InterledgerQuotingProtocol.asn"
```
sappenin: This was breaking Checkstyle...now fixed.

src/main/java/org/interledger/ilqp/QuoteByDestinationAmountResponse.java
```diff
@@ -4,7 +4,7 @@
 import java.util.Objects;
 
 /**
- * A quote sent in response to a request of type {@link QuoteByDestinationAmountresRequest}.
+ * A quote sent in response to a request of type {@link QuoteByDestinationAmountRequest}.
```
sappenin: This was breaking Checkstyle...now fixed.

sappenin:
Assuming I get a green build, I'm going to push this change. The only differences between what was already approved and what is in this PR now are as follows:

* Checkstyle fixes (Formatting, etc)
* Fixed a bug with `InterledgerAddress.startsWith` where the input is an InterlegerAddress.
* Change the result of `InterledgerAddress.toString`

Really, that last bullet is the only additional thing of substance - if anyone prefers to revert the toString change, let's just file an issue and discuss it in there.

src/main/java/org/interledger/InterledgerAddress.java
sappenin: Actually, while this is a trivial change, I'm going to move it to another PR - I feel icky sliding this into this PR without some discussion (even if it is a trivial thing).

src/test/java/org/interledger/InterledgerAddressTest.java
```diff
@@ -123,4 +135,9 @@ public void test_address_hashcode() {
     assertThat(addressPrefix1.toString().equals(addressPrefix3.toString()), is(false));
   }
 
+  @Test
+  public void testToString() throws Exception {
+    assertThat(InterledgerAddress.of("g.foo.bob").getValue().toString(), is("g.foo.bob"));
```
sappenin: This test passes, but is incorrect - I'll fix this in the next PR where I update the toString of `InterledgerAddress`.

src/test/java/org/interledger/InterledgerAddressTest.java
```diff
@@ -70,7 +82,7 @@ public void test_address_with_blank() {
    * Validates adding an address prefix (as an InterledgerAddress) to an address prefix.
    */
   @Test
-  public void test_address_prefix_with_address_prefix() {
```
sappenin: Don't think we run checkstyle on our unit tests, but I recall it not liking this underscore function naming style, so made these camel-case.

PR #68, opened 2017-08-04
feat: support liquidity quotes

andrew-g-za:
Hey chaps,

At last, the liquidity quote codec. I've also fixed a few things here and there, particularly our .toString convention of starting with 'Impl{..}' - _which_ Impl?

p.s. I had some weird merging issues, hopefully it doesnt get too confusing. If it does, i'll just check out the dev branch and drop my changes on top (since we squash + merge, it should be the same i think)


src/main/java/org/interledger/codecs/oer/ilqp/QuoteBySourceAmountRequestOerCodec.java
```diff
@@ -37,7 +37,7 @@ public QuoteBySourceAmountRequest read(CodecContext context, InputStream inputSt
         context.read(InterledgerAddress.class, inputStream);
     
     /* read the source amount, which is a uint64 */
-    /* NOTE: we don't expect amounts to exceed 2^63 - 1, so we risk the down-cast */
+    /* NOTE: we dont expect amounts to exceed 2^63 - 1, so we risk the down-cast */
```
adrianhopebailie: This seems like an odd change

src/main/java/org/interledger/ilqp/LiquidityPoint.java
```diff
@@ -0,0 +1,121 @@
...
+      private final BigInteger outputAmount;
+
+      private Impl(final Builder builder) {
+        Objects.requireNonNull(builder);
+
+        this.inputAmount =
+            Objects.requireNonNull(builder.inputAmount, "input amount must not be null!");
+
+        this.outputAmount =
+            Objects.requireNonNull(builder.outputAmount, "input amount must not be null!");
```
adrianhopebailie: "input amount must not be null!" => "output amount must not be null!"

src/main/java/org/interledger/ilqp/LiquidityPoint.java
```diff
@@ -0,0 +1,121 @@
...
+      return this;
+    }
+
+    /**
+     * Sets the output amount into the builder.
+     * 
+     * @param outputAmount An instance of {@link BigInteger}.
+     */
+    public Builder outputAmount(BigInteger outputAmount) {
+      this.outputAmount = Objects.requireNonNull(outputAmount);
```
adrianhopebailie: Do null checks in `build()`

src/main/java/org/interledger/ilqp/LiquidityPoint.java
```diff
@@ -0,0 +1,121 @@
...
+    private BigInteger inputAmount;
+    private BigInteger outputAmount;
+
+    /**
+     * Sets the input amount into the builder.
+     * 
+     * @param inputAmount An instance of {@link BigInteger}.
+     */
+    public Builder inputAmount(BigInteger inputAmount) {
+      this.inputAmount = Objects.requireNonNull(inputAmount);
```
adrianhopebailie: Do null checks in `build()`

src/main/java/org/interledger/ilqp/LiquidityCurve.java
```diff
@@ -0,0 +1,123 @@
...
+      return this;
+    }
+
+    /**
+     * Sets an individual liquidity point into the builder to be added to the curve.
+     * 
+     * @param liquidityPoint An instance of {@link LiquidityPoint}.
+     */
+    public Builder liquidityPoint(final LiquidityPoint liquidityPoint) {
+      Objects.requireNonNull(liquidityPoint);
```
adrianhopebailie: Do null checks in `build()`

src/main/java/org/interledger/ilqp/LiquidityCurve.java
```diff
@@ -0,0 +1,123 @@
...
+  class Builder {
+    private Set<LiquidityPoint> points = new HashSet<>();
+
+    /**
+     * Sets a liquidity curve into this builder.
+     * 
+     * @param curve An instance of {@link LiquidityCurve}.s
+     */
+    public Builder liquidityCurve(final LiquidityCurve curve) {
+      Objects.requireNonNull(curve);
```
adrianhopebailie: Do null checks in `build()`

src/main/java/org/interledger/ilqp/LiquidityCurve.java
```diff
@@ -0,0 +1,123 @@
...
+      this.points.clear();
```
adrianhopebailie: We should indicate in the comments that this clears all existing points.

src/main/java/org/interledger/ilqp/QuoteLiquidityResponse.java
```diff
@@ -48,5 +48,148 @@
...
+      return this;
+    }
+    
+    /**
+     * Set the applies-to address into this builder.
+     *
+     * @param appliesTo An instance of {@link InterledgerAddress}.
+     */
+    public Builder appliesTo(final InterledgerAddress appliesTo) {
+      this.appliesTo = Objects.requireNonNull(appliesTo);
```
adrianhopebailie: Do null checks in `build()`

src/main/java/org/interledger/ilqp/QuoteLiquidityResponse.java
```diff
@@ -48,5 +48,148 @@
...
+      return this;
+    }
+    
+    /**
+     * Set the source hold duration into this builder.
+     *
+     * @param sourceHoldDuration An instance of {@link Duration}.
+     */
+    public Builder sourceHoldDuration(final Duration sourceHoldDuration) {
+      this.sourceHoldDuration = Objects.requireNonNull(sourceHoldDuration);
```
adrianhopebailie: Do null checks in `build()`

src/main/java/org/interledger/ilqp/QuoteLiquidityResponse.java
```diff
@@ -48,5 +48,148 @@
...
+      return this;
+    }
+    
+    /**
+     * Set the expires-at into this builder.
+     * 
+     * @param expiresAt An instance of {@link ZonedDateTime}
+     */
+    public Builder expiresAt(final ZonedDateTime expiresAt) {
+      this.expiresAt = Objects.requireNonNull(expiresAt);
```
adrianhopebailie: Do null checks in `build()`

src/main/java/org/interledger/ilqp/QuoteLiquidityResponse.java
```diff
@@ -48,5 +48,148 @@
...
+    private Duration sourceHoldDuration;
+    private ZonedDateTime expiresAt;
+    
+    /**
+     * Set the liquidity curve into this builder.
+     *
+     * @param liquidityCurve To An instance of {@link LiquidityCurve}.
+     */
+    public Builder liquidityCurve(final LiquidityCurve liquidityCurve) {
+      this.liquidityCurve = Objects.requireNonNull(liquidityCurve);
```
sappenin: It's a minor style nit, but I prefer to perform null-checks as early as possible (like in these setter methods in the builder) because using null here is a bug - a developer should never null out a field in the builder because null is never valid. Conversely, general validation logic _should_ be in the build method because we don't want to validate potentially in-flight mutable values of the builder until 'build' is called, and then we do want to validate.

Al that to say, I vote for leaving the null checks here.

adrianhopebailie: The way I wanted to differentiate between failing early and failing during `build()` was based on whether the supplied object was mutable after being provided.

Thinking about that, it obviously still has to be non-null so I agree 👍 

Need to comb through other builders and ensure we're being consistent.

adrianhopebailie:
@andrew-g-za please squash and merge after fixing those two comments

src/main/java/org/interledger/codecs/oer/ilqp/QuoteBySourceAmountRequestOerCodec.java
andrew-g-za: oops, must be a merge fail. will revert

src/main/java/org/interledger/ilqp/QuoteLiquidityResponse.java
andrew-g-za: Yip agreed - it would be an error to try set a null liquidity curve into the quote response builder - the curve should always at least be empty rather than null. I'll leave this as-is.

src/main/java/org/interledger/ilqp/LiquidityPoint.java
andrew-g-za: Well spotted, will fix.

src/main/java/org/interledger/ilqp/LiquidityCurve.java
andrew-g-za: agreed, done.

PR #73, opened 2017-08-27
Fixes #57 (Transition Amount to BigDecimal)

### LINKED ISSUE #57: Switch to BigInteger from long
We are still using a long in many places for amounts. These are encoded as UInt64 which is too big for a long so there is a risk that we'll break things.

Need to switch to BigInteger and stop using deprecated functions on the UInt64 codecs.

**NOTE**: Need to ensure that we encode and decode accurately as the default byte encoding of BigInteger includes a sign bit. We only want unsigned 64 bit integers.

sappenin:
* Transitions all java-ilp-core “amount” fields to be BigInteger.
* Fix unit tests and Checkstyle errors.
* Increment version to 0.7.0 due to interface breaking change.


adrianhopebailie:
> Transitions all java-ilp-core “amount” fields to be BigDecimal

I think you mean BigInteger?

LGTM

src/main/java/org/interledger/codecs/oer/ilp/InterledgerPaymentOerCodec.java
```diff
@@ -47,9 +48,8 @@ public InterledgerPayment read(final CodecContext context, final InputStream inp
 
     // 3. Read the destinationAmount, which is a UInt64.
     /* NOTE: we don't expect amounts to exceed 2^63 - 1, so we risk the conversion to long */
-    long destinationAmount = context.read(OerUint64.class, inputStream)
-        .getValue()
-        .longValue();
+    final BigInteger destinationAmount = context.read(OerUint64.class, inputStream)
```
andrew-g-za: minor nit - the comments no longer apply, since we dont cast to long

src/main/java/org/interledger/ilqp/QuoteByDestinationAmountRequest.java
```diff
@@ -67,82 +66,81 @@ public Builder destinationHoldDuration(final Duration destinationHoldDuration) {
...
+       *                instances.
        */
       private Impl(final Builder builder) {
         Objects.requireNonNull(builder);
-        
+
         this.destinationAccount = Objects.requireNonNull(builder.destinationAccount,
             "destinationAccount must not be null!");
-        
-        if (builder.destinationAmount < 0) {
```
andrew-g-za: is there a compelling reason to remove the < 0 check? We cant put negative numbers on the wire, so it seems like a safe thing to do?

src/main/java/org/interledger/ilqp/QuoteByDestinationAmountResponse.java
```diff
@@ -69,20 +67,18 @@ public QuoteByDestinationAmountResponse build() {
...
 
       private Impl(final Builder builder) {
         Objects.requireNonNull(builder);
 
+        this.sourceAmount = Objects
+            .requireNonNull(builder.sourceAmount, "sourceAmount must not be null!");
         this.sourceHoldDuration = Objects.requireNonNull(builder.sourceHoldDuration,
             "sourceHoldDuration must not be null!");
 
-        if (builder.sourceAmount < 0) {
```
andrew-g-za: same question about the < 0 check.

src/test/java/org/interledger/codecs/oer/ilqp/IlqpCodecTests.java
```diff
@@ -31,6 +25,12 @@
 import java.time.ZonedDateTime;
 import java.time.temporal.ChronoUnit;
 
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+
```
andrew-g-za: did this pass the checkstyle checks? I seem to recall its very particular about import orders?

src/main/java/org/interledger/ilqp/QuoteByDestinationAmountRequest.java
sappenin: Per our discussion in the Java conference call, this has been restored.

PR #85, opened 2017-09-21
Merge development into master

adrianhopebailie:
Seems like a good time to do this


adrianhopebailie:
@sappenin bump...

sappenin:
I think we should wait until #87 is merged, no?

adrianhopebailie:
> I think we should wait until #87 is merged, no?

Actually I think we should merge and tag this, then we can start implementing the Immutables across all the objects

PR #86, opened 2017-09-22
[WIP] Add BTP codecs

adrianhopebailie:
Todo:
- [ ] Unit tests
- [ ] Test vectors
- [ ] Clean up CheckStyle

### DIFF
--- a/src/main/java/org/interledger/btp/BilateralTransferProtocolError.java
+++ b/src/main/java/org/interledger/btp/BilateralTransferProtocolError.java
@@ -1,0 +1,198 @@
+package org.interledger.btp;
+
+import java.time.ZonedDateTime;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+
+public interface BilateralTransferProtocolError extends BilateralTransferProtocolPacket {
+
+  BilateralTransferProtocolErrorTypes getError();
+
+  ZonedDateTime getTriggeredAt();
+
+  byte[] getData();
+
+  /**
+   * Get the default builder.
+   *
+   * @return a {@link Builder} instance.
+   */
+  static Builder builder() {
+    return new Builder();
+  }
+
+
+  /**
+   * A builder for instances of {@link BilateralTransferProtocolError}.
+   */
+  class Builder {
+
+    private long requestId;
+    private BilateralTransferProtocolErrorTypes errorType;
+    private ZonedDateTime triggeredAt;
+    private byte[] data;
+    private List<SubProtocolData> subProtocolData;
+
+    /**
+     * Set the request id
+     *
+     * @param requestId An UInt32 request id.
+     */
+    public Builder requestId(final long requestId) {
+      this.requestId = requestId;
+      return this;
+    }
+
+    /**
+     * Set the type.
+     *
+     * @param errorType A {@link BilateralTransferProtocolErrorTypes}
+     */
+    public Builder errorType(final BilateralTransferProtocolErrorTypes errorType) {
+      this.errorType = errorType;
+      return this;
+    }
+
+
+    /**
+     * Set the time that the error was triggered.
+     *
+     * @param triggeredAt A {@link ZonedDateTime}
+     */
+    public Builder triggeredAt(final ZonedDateTime triggeredAt) {
+      this.triggeredAt = triggeredAt;
+      return this;
+    }
+
+
+    /**
+     * Set the additional data.
+     *
+     * @param data A byte array
+     */
+    public Builder data(final byte[] data) {
+      this.data = data;
+      return this;
+    }
+
+    /**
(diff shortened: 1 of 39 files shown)


src/main/java/org/interledger/btp/BilateralTransferProtocolPrepare.java
```diff
@@ -0,0 +1,238 @@
...
+   * @return An instance of {@link Condition}.
+   */
+  Condition getExecutionCondition();
+
+  /**
+   * Expiry of the transfer.
+   *
+   * @return An instance of {@link ZonedDateTime}.
+   */
+  ZonedDateTime getExpiresAt();
```
sappenin: This should be a `java.time.Instant` to avoid Timezone confusion.

src/main/java/org/interledger/btp/BilateralTransferProtocolError.java
```diff
@@ -0,0 +1,198 @@
+package org.interledger.btp;
```
sappenin: Wow, our Interface pattern has a _TON_ of boiler-plate! In the absence of moving this project to Kotlin (ha-ha), we should seriously consider using the [Immutables](http://immutables.github.io/) library. 

Basically, you just define an interface, and then during the build phase the library will generate immutable implementations of each class, complete with builders that are _very_ nice to deal with. Even better, the dependency for Immutables has a `provided` scope, (the library is used only for compilation), so it's not really a dependency that anyone using the library would need to worry about.

It saves _so_ much time, and I've used it successfully in many production projects with no issues.

sappenin: As an example, this would be the implemenation for `BilateralTransferProtocolFulfilll`:

```
@Value.Immutable
public interface BilateralTransferProtocolFulfill extends BilateralTransferProtocolTransferPacket {
  Fulfillment getFulfillment();
}
```

That's it - free immutable implementation, free builder, extensible, smart/configurable (and free) equals+hashcode+toString, compatible with Jackson, etc, etc...

src/main/java/org/interledger/btp/BilateralTransferProtocolResponse.java
```diff
@@ -0,0 +1,129 @@
...
+    return new Builder();
+  }
+
+
+  /**
+   * A builder for instances of {@link BilateralTransferProtocolResponse}.
+   */
+  class Builder {
+
+    private long requestId;
```
sappenin: So, I think we might consider separating the `Packet` data from the payload. As an example, in the ASN the object called `BilateralTransferProtocolPacket` has a `requestId`, but the actual `Response` object has only subprotocol data.

In my implementation, I'm not going to be using `BilateralTransferProtocolPacket` because I won't be running on a Websocket, and thus won't need to track request/response identifiers for correlation. However, I _will_ want to re-use the Prepare/Fulfill/Reject/Message/Error interfaces. If we don't have a good separation between the `Packet` (i.e., encoded ASN.1) and the business objects, I won't be able to do that.

src/main/java/org/interledger/btp/BilateralTransferProtocolRuntimeException.java
```diff
@@ -0,0 +1,68 @@
...
+
+  /**
+   * Constructs a new BTP runtime exception with the specified detail message.
+   * The cause is not initialized, and may subsequently be initialized by a
+   * call to {@link #initCause}.
+   *
+   * @param message the detail message. The detail message is saved for
+   *                later retrieval by the {@link #getMessage()} method.
+   */
+  public BilateralTransferProtocolRuntimeException(String message) {
```
sappenin: I think we might want an optionally-present transferId here....most exception will have that, correct?

src/main/java/org/interledger/btp/SubProtocolContentType.java
```diff
@@ -0,0 +1,31 @@
+package org.interledger.btp;
+
+public enum SubProtocolContentType {
```
sappenin: I'm still not sure I like the restriction in content-type here...see discussion in [https://github.com/interledger/rfcs/issues/286](https://github.com/interledger/rfcs/issues/286) for more details. 

I suppose for now, this is the spec....

src/main/java/org/interledger/btp/SubProtocolData.java
```diff
@@ -0,0 +1,172 @@
...
+   * @return A String representing the protocol identifier.
+   */
+  String getProtocol();
+
+  /**
+   * The content type of the sub protocol data
+   *
+   * @return An instance of {@link SubProtocolContentType}.
+   */
+  SubProtocolContentType getContentType();
```
sappenin: Per my above comment, this should just be a String....?

adrianhopebailie:
Superseded by #87

PR #87, opened 2017-09-28
Add BTP codecs (using Immutables)

adrianhopebailie:
Replaces #86

### DIFF
--- a/src/main/java/org/interledger/btp/ErrorPacket.java
+++ b/src/main/java/org/interledger/btp/ErrorPacket.java
@@ -1,0 +1,16 @@
+package org.interledger.btp;
+
+import org.immutables.value.Value.Immutable;
+
+import java.time.Instant;
+
+@Immutable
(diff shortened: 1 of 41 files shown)


src/main/java/org/interledger/btp/ErrorPacket.java
```diff
@@ -0,0 +1,16 @@
+package org.interledger.btp;
+
+import org.immutables.value.Value.Immutable;
+
+import java.time.Instant;
+
+@Immutable
+public interface ErrorPacket extends BtpPacketBase {
+
+  BtpErrorType getError();
```
sappenin: Not a blocker, but I think Checkstyle will `warn` when Javadoc is missing and we should probably document each method in each interface.

adrianhopebailie: CheckStyle issues fixed

adrianhopebailie:
Migrating to Quilt repo

PR #95, opened 2017-10-15
Fixes #77

### LINKED ISSUE #77: Use java.time.Instant instead of ZonedDateTime
`java.time.Instant` will always be created using a fixed clock based off the epoch in UTC. We should consider moving to this object in order to enforce a consistent Timezone in our library.

For library users who want something more time-zone centric, Instant's can be trivially converted to ZonedDateTime or other java.time classes, like so:

```
Instant instant = Instant.now();
ZonedDateTime zdt = instant.atZone(ZoneId.of("PST"));
```

sappenin:
* Replace all instances of ZonedDateTime with Instant which always uses the UTC timezone.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 group = 'org.interledger'
 description = 'Core libraries for implementing the Interledger suite of protocols.'
-version = '0.8.3-SNAPSHOT'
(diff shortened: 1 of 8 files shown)


sappenin:
Yes, this is breaking - I'll bump the major version.

PR #99, opened 2017-10-17
96-fix-javadocs  

controlsurface:
I have updated the code to remove javadoc errors.


controlsurface:
@sappenin -- This is Chris Opler.  I will take on Add unit test coverage to CodecContext next !

Issue #88, opened 2017-09-29
Switch to using Maven only?


adrianhopebailie:
We currently have an overly complex build system with both Maven and Gradle.

The majority of contributors use Maven and it has proven to be better supported by plugins. It would be helpful to switch but I propose we take a contributor vote.

earizon:
Not problem for me to default to maven. I vote for the one that simplifies CI. (I'm not aware of the pros and cons of each one for CI)

adrianhopebailie:
I suggest we implement this change as part of migrating the code over to the Hyperledger Quilt project

rizwantanoli:
I am not a contributor yet, hope to become one, but my preference would be Maven

PR #103, opened 2017-10-18
Earizon shell scripts

earizon:
Added utility shell scripts for local development with common tasks/gradle recipes ("fast" gradle install, exec tests, full local install at this moment)


sappenin:
These look good, but given the commentary on #88, we'll be moving to maven so it's not clear that these scripts will be useful once that happens.

adrianhopebailie:
@earizon, can you tweak these to now use MVN and rebase on the latest development branch?

earizon:
@adrianhopebailie  Closing to avoid deprecated gradle commits and replacing with https://github.com/interledger/java-ilp-core/pull/108

PR #108, opened 2017-10-24
Earizon mvn script wrappers

earizon:
Utility shell scripts (shell wrappers around mvn command). Fast/Full local install and testing.


sappenin:
Sorry, can’t edit my comment but meant to say (at the end) that it might be worth adding something to the README in this folder about how best to use the scripts - but not a blocker at all.

adrianhopebailie:
@earizon can you move this PR to hyperledger/quilt

adrianhopebailie:
Moved to https://github.com/hyperledger/quilt/pull/34

